PR #49, opened 2021-01-26
Run bootstrap demo currency with parachain

clangenb:
Make client compatible with parachain

### DIFF
--- a/client/bootstrap_demo_currency.sh
+++ b/client/bootstrap_demo_currency.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-CLIENT="../target/release/encointer-client-notee -p 9946"
+CLIENT="../target/release/encointer-client-notee"
 
 # register new currency
 echo "registering demo currency with cid:"
@@ -28,10 +28,10 @@ account2=//Bob
 account3=//Charlie
 
 # charlie has no genesis funds
-#$CLIENT faucet $account3
+$CLIENT faucet $account3
 
-# await next block
-$CLIENT listen -b 1
+# wait long enough to make sure extrinsics are processed
+sleep 30
 
 $CLIENT --cid $cid register-participant $account1
 $CLIENT --cid $cid register-participant $account2
--- a/client/bootstrap_demo_currency.sh
+++ b/client/bootstrap_demo_currency.sh
@@ -43,7 +43,6 @@ sleep 30
 # list registry
 $CLIENT --cid $cid list-participants
 
-# Todo: for some reason it hangs here
 $CLIENT next-phase
 # should now be ASSIGNING
 
--- a/client/bootstrap_demo_currency.sh
+++ b/client/bootstrap_demo_currency.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-CLIENT="../target/release/encointer-client-notee"
+CLIENT="../target/release/encointer-client-notee -p 9946"
 
 # register new currency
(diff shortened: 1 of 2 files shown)


brenzi:
I don't like the timeout. What about waiting for two blocks?
```
$CLIENT listen -b 2
```

Are you sure Charlie is always endowed? also for old nodes?

clangenb:
> Are you sure Charlie is always endowed? also for old nodes?

No, he isn't, but I reverted that in 58c0be1

> I don't like the timeout. What about waiting for two blocks?
> 
> ```
> $CLIENT listen -b 2
> ```
I am not certain if it is deterministically 2 blocks, but I will check.

brenzi (approved):

client/src/main.rs
```diff
@@ -179,19 +179,9 @@ fn main() {
...
-                                GenericAddress::Id(to.clone()),
-                                PREFUNDING_AMOUNT
-                            )),
-                            nonce,
-                            Era::Immortal,
-                            _api.genesis_hash,
-                            _api.genesis_hash,
-                            _api.runtime_version.spec_version,
-                            _api.runtime_version.transaction_version
+                        let xt = _api.balance_transfer(
```
brenzi: nono, this won't work!
The faucet should be able to batch fund multiple accounts without always awaiting the next block. So we need to increase the nonce manually, as the api-client can't know the next nonce

brenzi (changes requested):

client/src/main.rs
clangenb: Ah yes, that totally makes sense. It didn't work, as the `Call::Balances` enum is wrong for the parachain case, which why I made the change in the first place, but I did not realize why it was implemented like that.

In that case, I will change it back and use the `compose_call!` macro instead.

brenzi: will this be fixed?
```                                                                                                                             
Faucet drips to 5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y (Alice's nonce=5)                                                                                                  
thread 'client' panicked at 'extrinsic error code 1010: Invalid Transaction: Transaction has a bad signature', /home/abrenzikofer/.cargo/git/checkouts/substrate-api-client-9327f54164a9792a/a6e1616/src/rpc/client.rs:173:33
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError', /home/abrenzikofer/.cargo/git/checkouts/substrate-api-client-9327f54164a9792a/a6e1616/src/lib.rs:454:45
```

clangenb:
I need to investigate. I suspect this is for the rebased parachain node?
Because that did not happen before.

brenzi:
yes, rebased parachain

client/bootstrap_demo_currency.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-CLIENT="../target/release/encointer-client-notee"
+CLIENT="../target/release/encointer-client-notee -p 9946"
```
brenzi: I don't like any parachain-specific things in here. The script should run with minimal CLI options and I don't want to modify tutorials

clangenb: I agree, which is why this was reverted in 4626f50cf112f47bcbe20d316f4bf8647dbaf876 :)

brenzi:
tested against encointer-parachain and encointer-node-notee. works

PR #50, opened 2021-02-11
[client] New command: get-proof-of-attendances

clangenb:


client/src/main.rs
```diff
@@ -623,6 +623,45 @@ fn main() {
...
+                    app.setting(AppSettings::ColoredHelp)
+                        .arg(
+                            Arg::with_name("accountid")
+                                .takes_value(true)
+                                .required(true)
+                                .value_name("SS58")
+                                .help("AccountId in ss58check format"),
+                        )
+                        .arg(
+                        Arg::with_name("amount")
```
brenzi: the naming "amount" seems odd to me.

People may not participate in subsequent ceremonies for the same cid. You can change location, you can skip ceremonies.

I would prefer to use the argument `ceremony-index` (`-i`) which can either be a positive absolute integer or a negative relative integer (-1 being the last ceremony) to precisely define a single ceremony index
Then you have to call this function for every proof, but that's nothing to worry about

brenzi (changes requested):

client/src/main.rs
```diff
@@ -623,6 +623,45 @@ fn main() {
...
+                    let cindex = get_ceremony_index(&api);
+                    let cid = verify_cid(
+                        &api,
+                     matches.value_of("cid").expect("please supply argument --cid"),
+                    );
+
+                    for i in 1..=amount {
+                        let proof = prove_attendance(accountid.clone(), cid, cindex - i, arg_who);
+                        info!("Proof: {:?}\n", &proof);
+                        println!("Proof: {:?}\n", hex::encode(proof.encode()));
```
brenzi: and please return nothing but a `0x` prefixed hex string, so we can easily parse this in a bash script

clangenb: Agreed, even command line copy-paste is currently a hassle

brenzi (approved):

PR #59, opened 2021-05-06
Client improvements

clangenb:
* add python client that wraps the rust client
* simplify bot-community.py with it
* rewrite bootstrap_demo_community in python. This change was motivated because it is way easier to parse process outputs in python. This is something we need for the ipfs support, when we want to extract the ipfs cid.


brenzi:
@nomelc should be aware of this!

clangenb:
Rebased & ready to merge

PR #87, opened 2021-06-25
[bazaar] add rpcs

NicJak:
[bazaar] add rpcs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1407,6 +1407,8 @@ dependencies = [
  "frame-benchmarking-cli",
  "jsonrpc-core",
  "log",
+ "pallet-encointer-bazaar-rpc",
+ "pallet-encointer-bazaar-rpc-runtime-api",
  "pallet-encointer-communities-rpc",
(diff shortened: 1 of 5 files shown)


brenzi:
@clangenb may I ask you to fix the build issues? I guess we just need to cargo update selectively, but I get
```
error: no matching package named `frame-benchmarking-cli` found
location searched: https://github.com/paritytech/substrate.git?branch=master#1d7f6e12
prerelease package needs to be specified explicitly
frame-benchmarking-cli = { version = "4.0.0-dev" }
required by package `encointer-node-notee v0.4.0 (/home/brenzi/encointer/encointer-node/node)`
```
is this related to https://github.com/integritee-network/integritee-node/issues/26 ?

clangenb:
Yes, this is related. Basically, with git dependencies, we always get a cargo update problem when substrate publishes a semver-incompatible version update because the version we reference does not exist at the newest git ref.

clangenb (approved):
Looks good to me!

PR #93, opened 2021-08-10
Add register query shop cli

clangenb:
#### Rust Client
* Add cli for `create-business`, `update-business`, `create-offering`, `list-business`, `list-offerings`, `list-business-offerings`. Closes #88.

#### Python Client
* Wrap all the new rust-cli cmds
* add a simple script that creates some random business/offering data and tests the new cli. Good base to complete #89.

#### CI
* add new integration-test: test-bazaar: Upload some business and offerings. But it does not verify the results yet.
* Add rust cache and reduce node build time to ~14 min

Note: the current implementation of the ipfs-wrapper does only support upload to the local ipfs deamon, and needs to be adjusted to upload to encointers pinned ipfs service. This should not be a big deal. I leave this to @anizeani.

Todo:
- [x] merge #92 first
- [x] rebase to master
- [x] retarget to master

PR #95, opened 2021-08-10
Support bazaar rpcs

clangenb:


clangenb:
Created this pr because the #93 did not trigger CI. It turns out that GA was just lagging.

PR #102, opened 2021-09-01
[client] Add remote ipfs functionality for bazaar

anizeani:
Uploading files to ipfs remote functionality added,
If flag -l or --ipfs-local is set, the local ipfs is used, else remote.
No more folder upload. 
Icons path as zip file uploaded, each business and offering as single json file added.

ipfs-api-key and the ipfs add to node api url are set as environment varibles. 

When getting file the filename is the hash, is it required to get the original name?


.github/workflows/build-and-test.yml
```diff
@@ -10,7 +10,7 @@ on:
 #    - cron:  '42 3 * * *'
 env:
   CARGO_TERM_COLOR: always
-
+
```
clangenb: please remove that.

client/py_client/ipfs.py
```diff
@@ -1,16 +1,23 @@
 import subprocess
 import re
 import warnings
+import os
 
-ICONS_PATH = '../test-data/icons'
+local_ipfs_key = None
 
+if os.path.isfile("client\py_client\config.py"):
```
clangenb: I don't get that. Why do you load that?

client/py_client/ipfs.py
```diff
@@ -1,16 +1,23 @@
...
+    print("local_ipfs_key: ", local_ipfs_key)
 
+ICONS_PATH = '../test-data/icons'
 class Ipfs:
     """ Minimal wrapper for the ipfs cli """
-
     @staticmethod
     def add_recursive(path_to_files):
-        ret = subprocess.run(["ipfs", "add", "-rw", path_to_files], stdout=subprocess.PIPE)
+        ret = subprocess.run(
```
clangenb: This is probably an automatic formatting issue. Can you make sure that you have linewraps at 120?

client/py_client/ipfs.py
```diff
@@ -29,3 +36,28 @@ def add_recursive(path_to_files):
...
+    @staticmethod
+    def add_recursive_remote(path_to_files, IPFS_API_KEY):
+        if IPFS_API_KEY != '':
+            ret = subprocess.run(
+                ["curl", "-X", "POST", "-F", f"file=@{path_to_files}", IPFS_API_KEY, "https://ipfs.infura.io:5001/api/v0/add"], stdout=subprocess.PIPE)
+        elif local_ipfs_key != None:
+            ret = subprocess.run(
+                ["curl", "-X", "POST", "-F", f"file=@{path_to_files}", local_ipfs_key, "https://ipfs.infura.io:5001/api/v0/add"], stdout=subprocess.PIPE)
+        # last line contains the directory cid
+        last = ret.stdout.splitlines()[-1]
```
clangenb: As the regex code is the same as mine, please extract it to a function and re-use it.

client/py_client/ipfs.py
```diff
@@ -29,3 +36,28 @@ def add_recursive(path_to_files):
     @staticmethod
     def add_recursive_multiple(paths):
         return [Ipfs.add_recursive(f) for f in paths]
+
+    @staticmethod
+    def add_recursive_remote(path_to_files, IPFS_API_KEY):
+        if IPFS_API_KEY != '':
+            ret = subprocess.run(
+                ["curl", "-X", "POST", "-F", f"file=@{path_to_files}", IPFS_API_KEY, "https://ipfs.infura.io:5001/api/v0/add"], stdout=subprocess.PIPE)
```
clangenb: The URL should not be hard-coded, please make an additional CLI.

clangenb (commented):
Looks in general good, but I have some comments. :)

clangenb:
I am not certain If I understand all your comments from this PR, please try to be more clear.

* What exactly is different with the curl command compared to the local one?
* When retrieving remote files with curl, you retrieve it with the CID. And when you get the files, the name is a hash, not the original name? I think you might be able to add a flag defining the name. But I am not sure. However, we don't need the name anyhow. This is not a problem.

The uploads are quite defined:
* Icons: They are needed by the app, and should be uploaded according to the structure you find in test-data
* The rest (businesses/offerings) are individual files containing jsons.

> Further: Should we use following structure?
Do you mean local folder structure? We can think about that, but then we'd also need per community icons and stuff.

client/py_client/ipfs.py
anizeani: @brenzi  should the url be interchangeable? and entered per cli?

.github/workflows/build-and-test.yml
anizeani: done

client/py_client/ipfs.py
anizeani: it was to use the ipfs_api_key locally from a config file. now i export the env variable and removed the logic for reading the key from a local file. -> done

client/py_client/ipfs.py
anizeani: done

anizeani: even though i just figured out to turn off totally formatting after save, which is undesired.. but I'll try to find out how to just not add new lines^^

client/py_client/ipfs.py
anizeani: i didn't manage this one yet, i uploaded a refactored method called refactoredCID, but if I refactor, I get an error message from the node: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: null, expected a string", line: 0, column: 0)', client/src/main.rs:384:26
I will try to solve it tomorrow morning

anizeani: now it works. eventhough the error message is not satisfying, the error was, that I wanted to create a new community with the gesell client and the ipfs_cid was None

anizeani: now the question would be: is it better to refactor the method to a static method or should I refactor to a method in the the helpers.py, which isn't part of the ipfs class, for example

client/py_client/ipfs.py
clangenb: Haha, it's better to fix your config. I admit, googling does not yield as straightforward results as you'd hope. Still, this page should have everything needed: https://code.visualstudio.com/docs/python/editing

client/py_client/ipfs.py
brenzi: I agree that the url should not be hardcoded. I'd rather have a cli argument for the url and another one for the api key. if both not present, use localhost, if api key not present, assume it isn'nt needed

brenzi: but, actually, the url and  api key should also be searched for in env variables, so you don't need to specify it explicitly each time. (override with cli should be possible though

brenzi:
please update the OP text. I believe your folder problem should be solved by one of rhese options:
• don't use folders but individual files, added separately
• use zip
both are acceptable for me. zip has performance benfits ,however

client/py_client/ipfs.py
clangenb: But what would I do if I want to to run the script against a local node? Then I'd need to overwrite the env var to be empty?

I don't have much practice with env vars, but I think I'd prefer having a CLI interface unless you say env vars is much better.

brenzi: we could do cli only, but this clutters the console with redundant info in manual use.
It is quite common to use env for such things. if you want to force localhost you can always override env with cli
My preference in this case is env+cli

anizeani: @brenzi so if I understand you correctly, I need you to add the following env variable:
IPFS_API_ADD: "https://ipfs.infura.io:5001/api/v0/add"

client/bot-community.py
```diff
@@ -61,13 +60,21 @@ def purge_keystore_prompt():
...
+    ipfs_cid = ''
+    if(ipfs_local):
+        ipfs_cid = Ipfs.add(zipped_folder)
+    else:
+        ipfs_api_key = os.environ['IPFS_API_KEY']
+        ipfs_add_url = os.environ['IPFS_ADD_URL']
+        ipfs_cid = Ipfs.add_remote(zipped_folder, ipfs_api_key, ipfs_add_url)
     print('initializing community')
-    b = init_bootstrappers(client)
+    b = init_bootstrappers(client)
```
clangenb: fmt

client/register-businesses.py
```diff
@@ -120,8 +135,12 @@ def shop_owners():
         client.await_block()
 
     create_offerings(cid, 5)
-    offerings_ipfs_cids = Ipfs.add_recursive_multiple(glob.glob(OFFERINGS_PATH + '/*.json'))
-    print(f'Uploaded offerings to ipfs: ipfs_cids: {offerings_ipfs_cids}')
+    if args.ipfs_local:
```
clangenb: We can save quite some lines if we consolidate the ipfs add methods to one method:

```
def add_multiple(files, local = false): # set default value of local to false.
      # extract env_vars if not local and use the remote version

      # use the local version if local.
   
```

clangenb (commented):
Looks good, and the bazaar is already working as it seems. Now you need to update the other integration test to use a zipped folder.

client/bot-community.py
anizeani: done

client/register-businesses.py
anizeani: done

anizeani: and false is not a reserved keyword in python, it's False ;)

clangenb (commented):
Looks good to me, but please change the file permissions back to 755. Otherwise, we can't execute the scripts with `./script.py` due to a lack of permissions.

clangenb (approved):
LGTM

PR #117, opened 2021-10-02
add icons to businesses and offerings

anizeani:
register businesses and offerings on node with image_cids on infura ipfs


brenzi (approved):
tested with bazaar-web. LGTM

PR #120, opened 2021-10-12
Add remote/local node or ipfs gateway options

anizeani:
this pr is no longer only about remote ipfs but also about remote chain connectivity.. maybe change name/create new branch?

- issues that I discovered: 
- ~"./bot-community.py benchmark" is stuck in infinity loop since commit [c3caf6e8e92a2db448f40079e62f996e6d886793](https://github.com/encointer/encointer-node/commit/c3caf6e8e92a2db448f40079e62f996e6d886793)~ no, just run phase.py or wait, until the defined time for next phase will occur (depending on the phase durations this can take very long time)
- ~once businesses are registered to a cid, it is not possible to register new businesses to the same cid. this is reproduced, by running the register-businesses.py script more then once. --> in rust code extrinsic error: ERROR encointer_client_notee] ExtrinsicFailed: Module { index: 12, error: 1, message: None } DispatchInfo { weight: 10000, class: DispatchClass::Normal, pays_fee: Pays::Yes }~ one account/owner has one business per cid. In future, private account can have multiple businesses as proxy. 
- running bot-community.py run throws: ERROR encointer_client_notee] ExtrinsicFailed: Other("") DispatchInfo { weight: 10000, class: DispatchClass::Normal, pays_fee: Pays::Yes }
also


clangenb:
I don't have precise insight on all issues because I did not have much to do with the recent updates. I try my best:

* `./bot-community.py benchmark` is intended to run forever, as far as I know. Do you think the behavior has changed?
* You can't register more than one business per owner per community. `update_business` call. 
https://github.com/encointer/encointer-node/blob/c25fb85ea1140a073a115bcbdd2b1275ad0056d2/client/src/main.rs#L803
* I think @brenzi knows that the scripts are failing now, but I am not sure.

brenzi:
as mentioned elsewhere: run phase.py alongside benchmark

brenzi (approved):
tested. works

PR #122, opened 2021-10-13
endorse account in bot-community.py

### LINKED ISSUE #62: add endorsements in bot-community.py script
One can now kickstart community growth with the addition of endorsements. This speeds up the benchmarking process.

- [ ] Adapt bot-community.py script
- [ ] Test stresstest bot-image from @nomelc  to work with new bot-community.py script

anizeani:
closes #62


client/src/main.rs
```diff
@@ -790,6 +790,50 @@ fn main() {
                     println!("{:?}", offerings);
                     Ok(())
                 }),
+        ).add_cmd(
+            Command::new("endorse-newcomer")
+                .description("endorses a newcomer")
+                .options(|app| {
+                    app.setting(AppSettings::ColoredHelp)
+                    .account_arg()
```
clangenb: account will be the sender of the extrinsic. you need to an additional argument, which will be the endorsee account

client/src/main.rs
```diff
@@ -790,6 +790,50 @@ fn main() {
...
+                    let proof = match rep {
+                        Reputation::Unverified => None,
+                        Reputation::UnverifiedReputable => None, // this should never by the case during REGISTERING!
+                        Reputation::VerifiedUnlinked => Some(prove_attendance(accountid, cid, cindex - 1, arg_who)),
+                        Reputation::VerifiedLinked => panic!("reputation of {} has already been linked! Not registering again", accountid),
+                    };
+                    debug!("proof: {:x?}", proof.encode());
+                    if get_current_phase(&api) != CeremonyPhaseType::REGISTERING {
+                        println!("wrong ceremony phase for registering participant");
+                        return Ok(())
```
clangenb: you don't need this part.

client/src/main.rs
```diff
@@ -790,6 +790,50 @@ fn main() {
...
+                    }
+                    let _api = api.clone().set_signer(sr25519_core::Pair::from(signer.clone()));
+                    let xt: UncheckedExtrinsicV4<_> = compose_extrinsic!(
+                        _api.clone(),
+                        "EncointerCeremonies",
+                        "register_participant",
```
clangenb: this should be endorse_newcomer

client/src/main.rs
```diff
@@ -790,6 +790,50 @@ fn main() {
...
+                        cid,
+                        proof
```
clangenb: this should be the endorsee then

brenzi:
is this ready? please mark PR ready and request my review

anizeani:
Endorsements implemented in run method. 
Endorsments work, but in await_block error is thrown: 
ExtrinsicFailed: Module { index: 9, error: 0, message: None } DispatchInfo { weight: 10000, class: DispatchClass::Normal, pays_fee: Pays::Yes }

anizeani:
> Endorsements implemented in run method. Endorsments work, but in await_block error is thrown: ExtrinsicFailed: Module { index: 9, error: 0, message: None } DispatchInfo { weight: 10000, class: DispatchClass::Normal, pays_fee: Pays::Yes }

I think this happens, if accounts are registered and not purged, and then registered again. --> run chain again with --tmp argument

brenzi:
ok, I'll try to reproduce. can you please resolve the conversations above if you consider them addressed?

brenzi:
1. I don't like your approach of creating, funding and endorsing accounts one-by-one. This takes way too long and is rather ugly compared to how the the rest of the script works. please do it the same way newbies are created (actually, I suggest to just increase the number of newbies to create by the number of endorsees in the current round and then you just need to specifically endorse the first N newbies with a single function call with the list of endorsees as argument. Less LOC and I'm pretty sure this will solve your issue mentioned above which probably is a race condition.
2. You continue endorsing accounts without taking note of how many endorsements are left. This leads to many failed extrinsics once the budget is exhausted. not nice for debugging

brenzi (changes requested):
see comment above

anizeani:
1. I implemented the methodology the way you suggested. 
2. I didn't understand where we can find out the contingent of available endorsees... So how do I find out how many are allowed?

clangenb (changes requested):
Why do you resolve all my comments without fixing? You clearly have not touched the issues I mentioned

brenzi:
@anizeani: The number of endorsees allowed is a runtime constant. You could fetch it through rpc, but I think it's fine to just define a constant in the py script. I believe we initialized the total number of endorsees per bootstrapper to 50

client/src/main.rs
```diff
@@ -1501,6 +1519,21 @@ fn endorse_newcomer(
 	Ok(())
 }
 
+fn get_BurnedBootstrapperNewbieTickets(
+	api: &Api<sr25519::Pair, WsRpcClient>,
+    bootstrapper: &AccountId,
+	key: CommunityCeremony
+) -> u8 {
+    let bootstrapper = matches.bootstrapper_arg().map(get_pair_from_str).unwrap();
+	api.get_storage_double_map(
```
clangenb: Your code should not compile because you don't pass matches here.

brenzi:
what is the status here?

clangenb:
When testing, there were nonce errors, which makes sense as multiple accounts are endorsed withing the same block, but we were using the online `compose_extrinsic`. However, we want to be able to endorse multiple endorsees at once to speed up the process. This means we need to manually increment the nonce either on python  or rust side. Regardless of the approach, the rust code needs to be touched, which means it hangs on me.

brenzi:
I suggest to do it like the cli `faucet` command in rust: count up nonces in the rust CLI. pretty much copy-paste

clangenb:
I decided to see this through to the end because it expected some ping-pong between rust and python adjustments (which did indeed happen).

* Loosely tested endorsements, seems to work.
* I was wondering about the statically assigned value of performing two endorsements per ceremony. This seems really low to me. Is this something you discussed @brenzi, @anizeani? Except for this value, I think this is ready to be merged.

anizeani:
If I'm not mistaken, then 2 endorsees is what we decided, but since this parameter is parametrizable, we can for stress test just augment it, right?

clangenb (approved):
I tested some more, added 101 endorsees, 2 newbies, could register them all successfully, for the first (bootstraping) ceremony. Which leads to the expected meetup distribution:

```
number of known accounts: 113
meetups assigned for 113 participants with sizes: [12, 12, 12, 11, 11, 11, 11, 11, 11, 11]
```

clangenb:
@brenzi ready to merge IMO.

client/bot-community.py
```diff
@@ -13,30 +13,34 @@
...
-from py_client.client import Client, ExtrinsicFeePaymentImpossible, ExtrinsicWrongPhase, UnknownError, ParticipantAlreadyLinked
+from py_client.client import Client, ExtrinsicFeePaymentImpossible, ExtrinsicWrongPhase, UnknownError, \
+    ParticipantAlreadyLinked
 from py_client.ipfs import Ipfs, ICONS_PATH
 from py_client.communities import populate_locations, generate_community_spec, meta_json
 
 KEYSTORE_PATH = './my_keystore'
 NUMBER_OF_LOCATIONS = 100
 MAX_POPULATION = 12 * NUMBER_OF_LOCATIONS
+NUMBER_OF_ENDORSEMENTS_PER_REGISTRATION = 101
```
clangenb, line 36: I reckon we will fine tune this afterwards. The script does not support meetups anyhow until we integrated handling of randomized meetups.

brenzi (approved):

PR #123, opened 2021-10-13
Cargo fmt and enforce it in ci

clangenb:


clangenb:
@brenzi FYI

clangenb:
@brenzi FYI

PR #128, opened 2021-10-27
collect ksm fees 18

NicJak:
please verify that the constants set are as intentioned by the task, as I am not proficient in this area.
I mapped the following:
"no block author reward. 100% to treasury" -> TREASURY_FUNDING_PERCENT
"no burn" -> pub const Burn: Permill = Permill::zero();

also: 
please propose the correct value for runtime/.../lib.rs:g317:
pub const ProposalBondMinimum: Balance = 1; //TODO add minimum balance

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -914,7 +914,7 @@ dependencies = [
  "log",
  "regalloc",
  "serde",
- "smallvec 1.6.1",
+ "smallvec 1.7.0",
  "target-lexicon",
 ]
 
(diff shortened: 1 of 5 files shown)


NicJak:
ps: the changes in the lock files come from one of the issues when tackling this task: there were issues with the compatibility of the encointer-client-notee package that were resolved this way with the help of @clangenb

brenzi:
what issue are you working on? Why do you add a treasury to the solochain?

brenzi:
you're in the wrong repo mate ;-) @NicJak

clangenb:
I did this in the parachain repo already ;).

So we can close this, and I don't need to review?

NicJak:
yes, nevermind

PR #129, opened 2021-11-17
fixed file location

anizeani:


anizeani:
minor change, but I can't merge to master by myself @brenzi (authorization required)

clangenb (approved):
So the scripts assume to be run from the repository root dir?

anizeani:
No, we refactored the test-data location into the client folder #126. The scripts are still run in the client folder. This is the change

clangenb:
Ok, I see. Thanks for explaining. 👍

PR #134, opened 2021-11-24
[runtime/#54] add pallet proxy

NicJak:


clangenb (approved):
Finally. :)

But you need to fix the fmt.

NicJak:
> Finally. :)
> 
> But you need to fix the fmt.

nooooooo...

PR #135, opened 2021-11-26
Use click for python cli parsing

### LINKED ISSUE #133: use click for python cli parsing
https://click.palletsprojects.com/en/8.0.x/

so much nicer than argparse

anizeani:
click framework is used instead of argparser
closes #133
Remarks: 
- for remote node, the url is hardcoded and the port also fixed. This is because I assume if we connect to remote we use tcp port 443 and wss://gesell.encointer.org. 
If you want I can make that generic, too. I just assumed that if remote people would choose this one in anycase.
- bootsrap_demo_community.py (on gesell 0.7.0): 
1st time excecutes correctly
2nd time throws an error:
ERROR encointer_client_notee] ExtrinsicFailed: Other("") DispatchInfo { weight: 10000, class: DispatchClass::Normal, pays_fee: Pays::Yes }
[2021-11-26T17:37:42Z ERROR encointer_client_notee] ExtrinsicFailed: Module { index: 10, error: 4, message: None } DispatchInfo { weight: 10000, class: DispatchClass::Normal, pays_fee: Pays::Yes }
this also happens on master branch. with 0.6.0 i didnt have error, so apparently breaks with 0.7.0 
- connection to remote node not possible ATM: faucet listens to localhost:5000/api. is there also a remote faucet_url or else i can't create the account on remote node. 
- ~in register-businesses.py: ifps_local had to be added as function argument in multiple function levels to reach the lowest point. This is because now we don't use args (args are global). Maybe I should use a global variable to solve that (other suggestion?)~ solved with global variable

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -49,15 +49,21 @@ listen to chain events for debugging (i.e. see failed extrinsics)
 RUST_LOG=encointer_client_notee=info ./target/release/encointer-client-notee listen
 ```
 
+run a single cycle
+```bash
+cd client
+./bot-community.py run
+```
+
 benchmark bot community
 ```bash
 cd client
-./bot-community.py init
+./bot-community.py benchmark
 ```
 
 if you'd like to test bazaar with dummy businesses and offerings too, you need to provide IPFS.
 
-either through infura:
+either through infura
 
 ```
 export IPFS_ADD_URL=https://ipfs.infura.io:5001/api/v0/add
@@ -66,7 +72,7 @@ export IPFS_API_KEY=<user>:<password>
 ./register-businesses.py
 ```
 
-or locally:
+or locally
 
 ```
 # you may need to run 'ipfs init'
@@ -74,3 +80,17 @@ ipfs daemon
 ./bot-communities.py init --ipfs-local
 ./register-businesses.py  --ipfs-local
 ```
+
+In IPFS, the community icons and data of businesses and offerings are stored.
+## Notes
+
+The cli provides helpful information for the ordering of options and commands. <br>
+You can use a custom chain client by providing the option before the command
+```
+./bot-communities.py --client ./path/to/custom/client init
+```
+You can also connect to a remote chain
+```
+./bot-communities.py init --node_url wss://remote.node.org
+```
+For now, the node_url is hardcoded to 'wss://gesell.encointer.org' despite which value you enter. This will be changed in the future when more remote chains will be available. 
--- a/client/README.md
+++ b/client/README.md
@@ -49,7 +49,7 @@ listen to chain events for debugging (i.e. see failed extrinsics)
 RUST_LOG=encointer_client_notee=info ./target/release/encointer-client-notee listen
 ```
 
-run a single cycle
+run a single phase
 ```bash
 cd client
 ./bot-community.py run
@@ -85,24 +85,12 @@ In IPFS, the community icons and data of businesses and offerings are stored.
 
 You can cat/get the data stored in ipfs locally:
 ```
-ipfs cat CONTENT_IDENTIFIER
+ipfs cat <CONTENT_IDENTIFIER>
 ```
 Or if it was stored remotely (on Infura):
 ```
-curl -X POST "https://ipfs.infura.io:5001/api/v0/cat?arg=QmaS7notdEkVmFvL2E19r8bpcVjdgjEV3gu6C2jYqUS1UB" 
+curl -X POST "https://ipfs.infura.io:5001/api/v0/cat?arg=<CONTENT_IDENTIFIER>" 
 ```
 ## Notes
 
-The cli provides helpful information for the ordering of options and commands. <br>
-The following options exist:
-* <kbd>--client</kbd> --> (string) path/to/chain/client
-* <kbd>--port</kbd> --> (string) specify port for the chain client
-* <kbd>-l,--ipfs_local</kbd> --> (bool) choose local ipfs node
-* <kbd>--node_url</kbd> --> (string) choose local ipfs node
-
-A possible call would look like
-
-```
-./bot-communities.py --client ./path/to/custom/client --port 123 -l --node_url wss://gesell.encointer.org init
-```
 For now, the node_url is hardcoded to 'wss://gesell.encointer.org' and port 443 is automatically set despite which value you enter. This will be changed in the future when more remote chains will be available. 
--- a/client/README.md
+++ b/client/README.md
@@ -69,7 +69,7 @@ either through infura
 export IPFS_ADD_URL=https://ipfs.infura.io:5001/api/v0/add
 export IPFS_API_KEY=<user>:<password>
 ./bot-communities.py init
-./register-businesses.py register
+./register-businesses.py
 ```
 
 or locally
@@ -78,10 +78,19 @@ or locally
 # you may need to run 'ipfs init'
 ipfs daemon
 ./bot-communities.py --ipfs-local init 
-./register-businesses.py --ipfs-local register
+./register-businesses.py --ipfs-local
 ```
 
 In IPFS, the community icons and data of businesses and offerings are stored.
+
+You can cat/get the data stored in ipfs locally:
+```
+ipfs cat CONTENT_IDENTIFIER
+```
+Or if it was stored remotely (on Infura):
+```
+curl -X POST "https://ipfs.infura.io:5001/api/v0/cat?arg=QmaS7notdEkVmFvL2E19r8bpcVjdgjEV3gu6C2jYqUS1UB" 
+```
 ## Notes
 
 The cli provides helpful information for the ordering of options and commands. <br>
--- a/client/README.md
+++ b/client/README.md
@@ -91,6 +91,3 @@ Or if it was stored remotely (on Infura):
 ```
 curl -X POST "https://ipfs.infura.io:5001/api/v0/cat?arg=<CONTENT_IDENTIFIER>" 
 ```
-## Notes
-
-For now, the node_url is hardcoded to 'wss://gesell.encointer.org' and port 443 is automatically set despite which value you enter. This will be changed in the future when more remote chains will be available. 
--- a/client/README.md
+++ b/client/README.md
@@ -69,7 +69,7 @@ either through infura
 export IPFS_ADD_URL=https://ipfs.infura.io:5001/api/v0/add
 export IPFS_API_KEY=<user>:<password>
 ./bot-communities.py init
-./register-businesses.py
+./register-businesses.py register
 ```
 
 or locally
@@ -77,20 +77,23 @@ or locally
 ```
 # you may need to run 'ipfs init'
 ipfs daemon
-./bot-communities.py init --ipfs-local
-./register-businesses.py  --ipfs-local
+./bot-communities.py --ipfs-local init 
+./register-businesses.py --ipfs-local register
 ```
 
 In IPFS, the community icons and data of businesses and offerings are stored.
 ## Notes
 
 The cli provides helpful information for the ordering of options and commands. <br>
-You can use a custom chain client by providing the option before the command
+The following options exist:
+* <kbd>--client</kbd> --> (string) path/to/chain/client
+* <kbd>--port</kbd> --> (string) specify port for the chain client
+* <kbd>-l,--ipfs_local</kbd> --> (bool) choose local ipfs node
+* <kbd>--node_url</kbd> --> (string) choose local ipfs node
+
+A possible call would look like
+
 ```
-./bot-communities.py --client ./path/to/custom/client init
+./bot-communities.py --client ./path/to/custom/client --port 123 -l --node_url wss://gesell.encointer.org init
 ```
-You can also connect to a remote chain
-```
-./bot-communities.py init --node_url wss://remote.node.org
-```
-For now, the node_url is hardcoded to 'wss://gesell.encointer.org' despite which value you enter. This will be changed in the future when more remote chains will be available. 
+For now, the node_url is hardcoded to 'wss://gesell.encointer.org' and port 443 is automatically set despite which value you enter. This will be changed in the future when more remote chains will be available. 
--- a/client/bootstrap_demo_community.py
+++ b/client/bootstrap_demo_community.py
@@ -54,8 +54,8 @@ def update_spec_with_cid(file, cid):
 
 
 @click.command()
-@click.option('--client', default='../target/release/encointer-client-notee', help='the client to communicate with the chain')
-@click.option('--port', default='9944', help='port for the client to communicate with chain')
+@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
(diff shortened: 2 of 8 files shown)


client/bootstrap_demo_community.py
```diff
@@ -54,7 +53,12 @@ def update_spec_with_cid(file, cid):
         spec_json.truncate()
 
 
-def main(ipfs_local, client=Client()):
+@click.command()
+@click.option('--client', default='../target/release/encointer-client-notee', help='the client to communicate with the chain')
+@click.option('--port', default='9944', help='port for the client to communicate with chain')
```
clangenb:
```suggestion
@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
@click.option('--port', default='9944', help='ws-port of the chain.')
```

clangenb: If possible, please put full sentences here.

client/README.md
```diff
@@ -49,15 +49,21 @@ listen to chain events for debugging (i.e. see failed extrinsics)
 RUST_LOG=encointer_client_notee=info ./target/release/encointer-client-notee listen
 ```
 
+run a single cycle
+```bash
+cd client
+./bot-community.py run
+```
```
clangenb: I think the comment here is not correct. It does only execute a phase, no?

client/README.md
```diff
@@ -66,11 +72,37 @@ export IPFS_API_KEY=<user>:<password>
...
+
+In IPFS, the community icons and data of businesses and offerings are stored.
+
+You can cat/get the data stored in ipfs locally:
+```
+ipfs cat CONTENT_IDENTIFIER
+```
+Or if it was stored remotely (on Infura):
+```
+curl -X POST "https://ipfs.infura.io:5001/api/v0/cat?arg=QmaS7notdEkVmFvL2E19r8bpcVjdgjEV3gu6C2jYqUS1UB"
```
clangenb: Consequently you should also put here:

```
 curl -X POST "https://ipfs.infura.io:5001/api/v0/cat?arg=<CONTENT_IDENTIFIER>" 
```

client/bot-community.py
```diff
@@ -36,6 +34,82 @@
...
+    write_cid(cid)
+
+
+@cli.command()
+@click.pass_obj
+def run(ctx):
+    return run_no_annotators(ctx['client'])
+
+
+def run_no_annotators(client: Client):
```
clangenb: I would suggest calling this function `execute_current_phase`

client/bot-community.py
```diff
@@ -36,6 +34,82 @@
 NUMBER_OF_LOCATIONS = 100
 MAX_POPULATION = 12 * NUMBER_OF_LOCATIONS
 
+
+@click.group()
+@click.option('--client', default='../target/release/encointer-client-notee', help='the client to communicate with the chain')
+@click.option('--port', default='9944', help='port for the client to communicate with chain')
+@click.option('-l', '--ipfs_local', is_flag=True, help='if set, local ipfs node is used')
+@click.option('--node_url', default=None, help='if set, remote chain is used with port 443, no need to manually set port, it will be ignored')
```
clangenb: The way you use this arg, I would suggest that you simply call it `--gesell` and make it a flag.

clangenb: Or even better, rename the flag to `--remote-chain`, and allow only the value `gesell` now.

client/py_client/helpers.py
```diff
@@ -52,4 +54,12 @@ def take_only_last_cid(ret_cids):
...
-            return ''
\ No newline at end of file
+            return ''
+
+
+def set_local_or_remote_chain(client: str, port: str, node_url: str):
+    if (node_url == None):
+        client = Client(rust_client=client, port=port)
+    else:
+        client = Client(rust_client=client, node_url='wss://gesell.encointer.org', port=443)
```
clangenb: Dito, pass the flag here and check if its value is `gesell`, throw an error if it is an unrecognized flag.

client/README.md
```diff
@@ -66,11 +72,37 @@ export IPFS_API_KEY=<user>:<password>
...
+* <kbd>--port</kbd> --> (string) specify port for the chain client
+* <kbd>-l,--ipfs_local</kbd> --> (bool) choose local ipfs node
+* <kbd>--node_url</kbd> --> (string) choose local ipfs node
+
+A possible call would look like
+
+```
+./bot-communities.py --client ./path/to/custom/client --port 123 -l --node_url wss://gesell.encointer.org init
 ```
+For now, the node_url is hardcoded to 'wss://gesell.encointer.org' and port 443 is automatically set despite which value you enter. This will be changed in the future when more remote chains will be available.
```
clangenb: update the doc after changing `--node-url` to `--gesell`

client/README.md
```diff
@@ -66,11 +72,37 @@ export IPFS_API_KEY=<user>:<password>
...
+```
+## Notes
+
+The cli provides helpful information for the ordering of options and commands. <br>
+The following options exist:
+* <kbd>--client</kbd> --> (string) path/to/chain/client
+* <kbd>--port</kbd> --> (string) specify port for the chain client
+* <kbd>-l,--ipfs_local</kbd> --> (bool) choose local ipfs node
+* <kbd>--node_url</kbd> --> (string) choose local ipfs node
+
```
clangenb: I am unsure if we should list the options. This kind of information gets outdated easiliy.

clangenb (commented):
Some comments and suggestions, but it looks good in general 👍

client/README.md
anizeani: Yes, I'll change that

client/README.md
anizeani: True

client/bot-community.py
anizeani: We should discuss this. I left node_url a string instead of bool because I wondered, if we will want to add other url's. If yes, i can just make it directly generic and not hardvode the string.

anizeani: @brenzi do you wish the remote node url to be hardcoded or configurable?

clangenb: I think it makes sense to hard-code them, as I can only think of 4 encointer chains that will be run long-term remotely.

* gesell-kusama
* gesell-polkadot
* cantillon-kusama
* cantillon-polkadot

Short term we might have gesell-/cantillon-westen, but I doubt that these scripts should support that.

anizeani: do gesell-dot, cant ksm & cant dot exist? I don't see them on polkadot-js. if yes, what are the ws adresses?

clangenb: Ah sorry, I forgot the two obvious ones: gesell and cantillon standalone. These are the only ones that already run.

The ones from my previous comment don't exist yet.

anizeani: is for cantillon the address: 
wss://cantillon.encointer.org
?

clangenb: It would be, but it does not seem to be running, nor is it supported currently. Hence, I would not add it for now.

client/bot-community.py
```diff
@@ -36,6 +34,83 @@
 NUMBER_OF_LOCATIONS = 100
 MAX_POPULATION = 12 * NUMBER_OF_LOCATIONS
 
+
+@click.group()
+@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
+@click.option('--port', default='9944', help='ws-port of the chain.')
+@click.option('-l', '--ipfs_local', is_flag=True, help='if set, local ipfs node is used')
+@click.option('-r', '--remote_chain', default=None, help='choose one of the remote chains: gesell, gesell-dot, gesell-ksm, cantillon-dot, cantillon-ksm')
```
clangenb: Here I would only put the current possible remote chains. Hence, only `gesell`

client/bot-community.py
```diff
@@ -36,6 +34,83 @@
...
+    print(f'created community with cid: {cid}')
+    write_cid(cid)
+
+
+@cli.command()
+@click.pass_obj
+def run(ctx):
+    return execute_current_phase(ctx['client'])
+
+
```
clangenb: 1. Do you have any use case for this command? I am not sure if it is useful.
2. I think the cmd should have a name similar to the function

client/register-businesses.py
```diff
@@ -1,23 +1,77 @@
...
+global IPFS_LOCAL
+
+@click.command()
+@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
+@click.option('--port', default='9944', help='ws-port of the chain.')
+@click.option('-l', '--ipfs_local', is_flag=True, help='if set, local ipfs node is used.')
+@click.option('-r', '--remote_chain', default=None, help='choose one of the remote chains: gesell, gesell-dot, gesell-ksm, cantillon-dot, cantillon-ksm.')
+def register(client, port, ipfs_local, remote_chain):
+    client = set_local_or_remote_chain(client, port, remote_chain)
+    global IPFS_LOCAL
```
clangenb, line 28: You don't need to define a global twice. :)

client/phase.py
```diff
@@ -8,42 +8,28 @@
...
-if __name__ == '__main__':
-    p = argparse.ArgumentParser(
-    prog='register-businesses', parents=[simple_parser()])
-    args = p.parse_args()
-    # print(f"Starting script with client '{args.client}' on port {args.port}")
+@click.command()
+@click.option('-r', '--remote_chain', default=None, help='choose one of the remote chains: gesell, gesell-dot, gesell-ksm, cantillon-dot, cantillon-ksm')
+@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
+@click.option('--port', default='9944', help='ws-port of the chain.')
+def main(remote_chain, client, port):
```
clangenb, line 25: Do we need to name the `main` function when calling the script, .i.e,

`./phase.py main <args>`?

client/phase.py
```diff
@@ -53,3 +39,18 @@ def subscription_handler(event_count, update_nr, subscription_id):
         print('NEXT PHASE!')
         client.next_phase()
         COUNT = 0
+
+
+def subscription_handler(event_count, update_nr, subscription_id):
+    global COUNT
```
clangenb, line 45: You don't need to redefine the global here.

client/register-businesses.py
```diff
@@ -1,23 +1,77 @@
...
 ICON_PATH = './test-data/icons/community_icon.png'
 
+global IPFS_LOCAL
+
+@click.command()
+@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
+@click.option('--port', default='9944', help='ws-port of the chain.')
+@click.option('-l', '--ipfs_local', is_flag=True, help='if set, local ipfs node is used.')
+@click.option('-r', '--remote_chain', default=None, help='choose one of the remote chains: gesell, gesell-dot, gesell-ksm, cantillon-dot, cantillon-ksm.')
+def register(client, port, ipfs_local, remote_chain):
```
clangenb: I prefer the function name to be more specific: `register_businesses`. I wonder here too. As it is the only cmd in the script, can it be omitted when running the script?

clangenb (commented):
Still have some comments. :)

In general: Try to avoid globals whenever possible. It makes unit-testing extremely hard. I don't think it matters much here, but I'd prefer not to have them, I guess.

client/bot-community.py
anizeani: yes,
I got stuck with the click framework and nested commands. 
The reason is, that if you call one function with click annotators within another (e.g. run within benchmark), you get a mess with the arguments of the function (you can try it out)
I used this solution here:
https://stackoverflow.com/a/40094408
But I renamed it now to run_without_annotators, because then run can be a slim wrapper around it

anizeani: for 2. thats why i called it: "run_without_annotators" ^^ (you suggested to change it ;)

client/phase.py
anizeani: no, thats the thing, you don't need to add the command. there are two ways to excecute the script without to add a command with the click framework: 
1. by using if __name__ == __main__ (this is how we do it), and then we call a function, in which we excecute what ever we want to excecute without needing to use it as additional command in the cli (so we keep the application the same way it was before)
2. with setupt.py and define there inside where to start. 
we are used to __name__ ==__main__ so I just kept it the way it is, but we need a function 
because we want to generically change the client/port, we need a function with annotators. and we need to call the function... (the name is irrelevant in that case)

client/phase.py
anizeani: i think I do. if I dont do this, I get a "Unresolved reference 'COUNT' " in line 46 
global just tells the function to treat it as global and not as local variable. apparently you neet to do it in every function when you want to use the same variable. alternatively we could pass it with the subscription handler
https://stackoverflow.com/questions/23998471/global-variable-changed-by-multiple-functions-how-to-declare-in-python

client/register-businesses.py
anizeani: yes, you don't need to add the command in the cli, just ./register-businesses.py [OPTIONS]

client/register-businesses.py
anizeani: surprisingly i did

client/phase.py
clangenb: Perfect. This is what I hoped.

client/phase.py
clangenb: This is the weirdest concept I have ever heard, but it seems you are right.

client/bot-community.py
clangenb: What I wanted to ask is: Why do you expose this as a command?

```
@cli.command()
@click.pass_obj
def run(ctx):
```

But maybe I did not get your point. :)

anizeani: I agree with you on that question, that it is questionable if we want to expose this command. 
I just wanted to keep the logic as close as possible to what it was, i.e. the tests run
./bot-community.py run 
so i thought lets keep it exposed..

client/register-businesses.py
```diff
@@ -65,7 +119,8 @@ def random_business():
     :return:
     """
     print("adding business image to remote: ")
-    image_cid = Ipfs.add(ICON_PATH, args.ipfs_local)
+    print(f"IPFS IS: {IPFS_LOCAL}")
```
clangenb: This print statement is misleading because `IPFS_LOCAL`  is either true or false. The statement should be like `Ipfs is local: {IPFS_LOCAL}`

anizeani: this print statement should have been removed, I just forgot it because I wanted to make sure I have to declare the global variable twice.. forgot to remove it

client/bot-community.py
clangenb: Ah the tests need this. Hence, my suggestion:

```
@cli.command()
@click.pass_obj
def execute_current_phase(ctx):
    return _execute_current_phase(ctx['client'])
```

anizeani: ok, so I also adjust the tests

clangenb (approved):
Nice, looks good now.

PR #145, opened 2021-12-23
Infura upload folder structure

anizeani:
A while ago we uploaded the icon folder as zip file and sent it to infura, mainly because I didn't manage to figure out how to recursively add folder structure to ipfs via infura. 

the advantage of zipping files is, that you always send one file, this can simplify things on the uploading side.
on the other hand, the files extraction and handling on the downloading side is more complicated.

In this PR I removed the zipping, made the local ipfs adding recursive again and implemented for remote ipfs upload via infura a http client to upload a folder structure or single file

### DIFF
--- a/client/py_client/ipfs.py
+++ b/client/py_client/ipfs.py
@@ -23,7 +23,6 @@ class Ipfs:
             return "QmP2fzfikh7VqTu8pvzd2G2vAd4eK7EaazXTEgqGN6AWoD"
         if local:
             ret = subprocess.run(["ipfs", "add", "-rw", path_to_files], stdout=subprocess.PIPE)
-            print(ret)
             return take_only_last_cid(ret)
         else:
             headers = { }
@@ -43,7 +42,6 @@ class Ipfs:
             response = requests.post('https://ipfs.infura.io:5001/api/v0/add', headers=headers, params=params, files=files, auth=(auth[0], auth[1]))
 
             for line in response.text.split("\n"):
-                # print(line)
                 data = json.loads(line)
                 if os.path.isfile(path_to_files):
                     return data["Name"]
@@ -73,7 +71,6 @@ class Ipfs:
         else:
             rel_path = ''
             rel_path = Path(rel_path)
-            os.path.basename(path_to_files)
             with open(os.path.abspath(path_to_files), 'rb') as file:
                 args += [(rel_path.as_posix(), file.read())]
         return args
--- a/client/py_client/ipfs.py
+++ b/client/py_client/ipfs.py
@@ -27,21 +27,28 @@ class Ipfs:
             return take_only_last_cid(ret)
         else:
             headers = { }
-            params = (
-                ('pin', 'true'),
-                ('recursive', 'true'),
-                ('wrap-with-directory', 'true'),
-            )
+            params = ()
+            if os.path.isdir(path_to_files):
+                params = (
+                    ('pin', 'true'),
+                    ('recursive', 'true'),
+                    ('wrap-with-directory', 'true'),
+                )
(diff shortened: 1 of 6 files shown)


brenzi:
does this work with our own gateway too? Otherwise I won't approve

anizeani:
> does this work with our own gateway too? Otherwise I won't approve

It works with every gateway, took me a while to test, since our gateway was not responding for a while yesterday. Also tested on the app side. (Still like someone  to verify :)

client/py_client/ipfs.py
```diff
@@ -15,18 +17,43 @@
...
+                    ('recursive', 'true'),
+                    ('wrap-with-directory', 'true'),
+                )
+            else:
+                params = (
+                    ('pin', 'true'),
+                )
+            files = generate_file_list(path_to_files)
+            auth = ipfs_api_key.split(":")
+            response = requests.post('https://ipfs.infura.io:5001/api/v0/add', headers=headers, params=params, files=files, auth=(auth[0], auth[1]))
```
clangenb, line 41: Shouldn't we make this configurable to also use the encointer end-point?

clangenb (commented):
Looks good, except for my one remark.

client/py_client/ipfs.py
anizeani: No, this is how we add a file to ipfs using the infura api with the user -u argument.
When a file is added to an ipfs node, it will notify the other nodes that it has the content with cid xxx.
So uploading like this will still allow us to view the file with our encointer gateway.
I dont remember exactly why we actually use infura, i guess its because by specifying a user & password it uploads the files to our node,  and nobody can upload to our node without knowing our user&password. So we are protected from attacks that fill up our space on the node. And the reason why we use our own node is to guarantee that the files will remain online?

brenzi: there is too much confusion here about our ipfs strategy. Please leave as-is for now. We will fix this in January

anizeani: ok, but for the record, this would be a working PR that works also on the app side. without this, we don't fetch correctly any new icons

brenzi: we can merge this if @clangenb agrees. ipfs fixe can and should be its own PR

clangenb (approved):
I have the feeling some parts of the code is more complex than necessary, but I have no domain knowledge to give better suggestions. :)

I guess this one is obsolete then? https://github.com/encointer/encointer-wallet-flutter/pull/249

client/py_client/ipfs.py
anizeani: > I have the feeling some parts of the code is more complex than necessary, but I have no domain knowledge to give better suggestions. :)
> 
> I guess this one is obsolete then? [encointer/encointer-wallet-flutter#249](https://github.com/encointer/encointer-wallet-flutter/pull/249)

yes, this would make the other PR on the app side obsolete. 
And I agree, the implementation looks more complex then it should, and I hoped it would be easier applieable, but it didn't work the easy way. I believe the ipfs http client for python should be an abstraction level higher and a better wrapper for this implementation: https://github.com/ipfs-shipyard/py-ipfs-http-client, but when I tried it a cuple of months ago i didn't succeed and didn't try it ever since (maybe now I'd have sufficient knowledge to solve it, but I wouldn't invest time in it unless you guys want to). 
There is for infura a python tutorial, and they use the "requests" library (which I also use in this PR):  
https://blog.infura.io/getting-started-with-infuras-ipfs-service-updated-june-2021/

If you're interested to see how it would looks like to upload folder structure with the curl command, you can see the commented code in my first commit

Issue #118, opened 2021-10-02
[CI] validate community growth


brenzi:
currently, CI does not really grow a community. It just runs the registering phase, not even an entire ceremony.

Instead of running `bot-community.py run` we should run `benchmark` and introduce a flag `--ci` which runs 2 full ceremonies and then verifies the total issuance expected

ends with exit code >0 if total issuance is wrong

brenzi:
fixed in #159

PR #155, opened 2022-01-18
add `ceremonies_getReputation` rpc

clangenb:


clangenb:
@NicJak when you are ready, can you do the same thing for the following rpc PRs of yours?

runtime/src/lib.rs
```diff
@@ -595,6 +596,13 @@ impl_runtime_apis! {
 		}
 	}
 
+
+	impl pallet_encointer_ceremonies_rpc_runtime_api::CeremoniesApi<Block, AccountId> for Runtime {
+		fn get_reputations() -> Vec<(CommunityCeremony, AccountId, Reputation)> {
+			EncointerCeremonies::get_reputations()
```
brenzi, line 602: Why plural? Does this return reputation for all accounts? Why?

brenzi: after discussion with @clangenb I believe we should merge this as is and accept bad performance for now. The solution will be slightly more involved than I thought:
https://github.com/encointer/pallets/issues/118

brenzi (approved):

brenzi:
I doubt this has ever worked....how did you test it?
this should return reputation for a specified account, while the pallet only allows fetching reputation for all accounts. there is a mismatch currently, I believe

PR #168, opened 2022-02-21
Add files without folder with ipfs

anizeani:
i realized, that since we added recursively files to ipfs (including a folder), register businesses didn't work anymore properly.
so i added the functionality of adding single files (businesses/offerings).

I tested it for: 
local chain with 
1. remote ipfs (infura) 
2. local ipfs daemon
Unfortunately, I had a 403 rpc method forbidden error when working with localhost and ubuntu 20.04 (native) 
On windows 11 with wsl 2 it worked without additional configurations (if I didnt also add it long time ago, if it doesn't work, add the same configurations as described below).
I think the problem was related to the ipfs daemon, that wasn't configured and what I think helped me to solve it was:
- closing the ipfs daemon in the terminal.
- adding the following configurations to the daemon (in the terminal):
```
ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
ipfs config Addresses.API /ip4/0.0.0.0/tcp/5001
ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin "[\"*\"]"
ipfs config --json API.HTTPHeaders.Access-Control-Allow-Credentials "[\"true\"]"
```
3. starting the daemon

Maybe we should document it elsewhere, too.. it wasn't a straightforward solution..

### DIFF
--- a/client/py_client/ipfs.py
+++ b/client/py_client/ipfs.py
@@ -57,3 +57,22 @@ class Ipfs:
         return [Ipfs.add_recursive(f, local) for f in paths]
 
 
+    @staticmethod
+    def add(path_to_files, local=False):
+        if not (use_ipfs_gateway or local):
+            return "QmWgTp4fBkxyUhnMrx4UVVqQ2McTQKJzq8yq3J5tCzdtfx"
+        if local:
+            ret = subprocess.run(["ipfs", "add", path_to_files], stdout=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


brenzi (approved):
I believe ipfs.py needs some refactoring, but that would be out of scope.
I don't really like the handling of `use_ipfs_gateway` and `local`. But TBD later

PR #176, opened 2022-02-26
[CI] auto release with artifacts upon tagging

### LINKED ISSUE #101: [CI] publish node and client binary artifacts upon release
CI artifacts should be directly added to new releases

please study this workflow which does it all:
https://github.com/integritee-network/parachain/blob/master/.github/workflows/ci.yml

brenzi:
Closes #101

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -386,7 +386,7 @@ jobs:
   publish-runtimes:
     name: Publish Runtimes
     runs-on: ubuntu-latest
-    needs: [release]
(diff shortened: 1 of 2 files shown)


brenzi:
@clangenb srtool seems to somehow mess with std as it expects `Serialize` in the build. do you have a clue why this CI doesn't pass?

clangenb:
I am investigating. Seems to be a feature flag issue. I get the same when I build the runtime's standalone locally.

clangenb:
* Fixed in #177

PR #178, opened 2022-02-28
[client] add `ceremony_index_arg` for meetups stats commands

clangenb:
For better meetup stats browsing.


client/src/cli_args.rs
```diff
@@ -85,7 +85,6 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
 			Arg::with_name(CEREMONY_INDEX_ARG)
 				.takes_value(true)
 				.allow_hyphen_values(true)
-				.default_value("-1")
```
clangenb, line 88: Defaulting to `-1` made only sense when it was used to create a `ProofOfAttendance`. I changed it there to an `unwrap_or(-1)`.

brenzi (approved):
really nice for debugging, thx!

PR #180, opened 2022-02-28
Fix upload url

mosonyi:


brenzi:
I would be very surprised if this would work because no job depends on build-runitmes. I think for integritee you had a `create_artifacts` job which depends on `build-runtimes` but is missing here. May I ask you to check again?

mosonyi:
It will work, because the `build-runtimes` uploads the artifacts ( `actions/upload-artifact@v2` ), which will be used by the `release `

brenzi:
I'm pretty sure race conditions can occur if build-runtimes takes longer. but let's merge and see

brenzi (approved):

PR #189, opened 2022-03-15
allow 1/3 newbies ratio

brenzi:
newbie divider was wrongly set to 4 instead of 3. will set it even to 2 to allow more newbies on testnet 1/3
fixing some CI stuff


clangenb (approved):
Looks good to me, but I don't understand why it should break the CI.

brenzi:
Because it changes community growth.
But probably we hardcoded the newbie ratio in bot-communities so we may need to make ut break before fixing

clangenb:
Yes, it is hard-coded:

 https://github.com/encointer/encointer-node/blob/2bc883d99fd0449f38bd5e4cce6eab43ee238817/client/bot-community.py#L223-L227

clangenb (approved):
LGTM

PR #190, opened 2022-03-19
Integrate encointer benchmarks and add weights

clangenb:
* For the time being, I added substrate's default weights for native pallets. At the most this might slightly more expensive than necessary because the substrate node's runtime is more complex than ours.


brenzi (approved):
LGTM

PR #192, opened 2022-03-19
add cli command to purge community-ceremony

### LINKED ISSUE #191: declutter reputation history on Gesell
today we changed from a 30min to a 4d ceremony cycle.
The current ceremony index is 1593

Almost all reputation history was preserved because we wanted reputation to survive at least 1-2 weeks

now, the rpc query for reputation takes 24s without caching. 

better to delete history with our new dispatchable
https://github.com/encointer/pallets/blob/f37679a922a675f7a98a9006704d32fe6c3874ec/ceremonies/src/lib.rs#L451

preserve these cindexes

1056, u0qj94fxxJ6, Reputation::VerifiedUnlinked
384, u0qj94fxxJ6, Reputation::VerifiedUnlinked
486, u0qj94fxxJ6, Reputation::VerifiedUnlinked

384, u0qj94fxxJ6, Reputation::VerifiedUnlinked
532, u0qj94fxxJ6, Reputation::VerifiedUnlinked

need to implement a client call for the purge command as a batch and output encoded call to submit as sudo

* [x] measure rpc respons time (28-33s for `time ./encointer-client-notee -u wss://gesell.encointer.org -p 443 reputation 5HEbRQGjxxjbGqN1JYwfxzYfSmNHiuRnwHZp2QdjjrPmGTx6`)
* [x] purge bot and test communities history `[1..1590]`: `e5dvt5mjcem`, `srcq45PYNyD`, `u0qj95PYNyD`(turbinenplatz. history might linger around)
   * 1590 would have exhausted weight limit. 
   * but for e5dvt5mjcem also 1..800 drove block time to >1min at execution and 801..1590 took 10min. and then after that it takes the api node again 10min to process the block. So, weight for the purge call should be multiplied with 100 at least
   * this community was stressed by a bot community, so the registry was well populated
   * only removing e5dvt5mjcem reduced rpc response time to 1.7s!!!!
   * removing srcq45PYNyD reduced it further to 0.9s
* [x] set reputation lifetime to 1500
* [x] measure rpc respons time
* [x] purge EdisonPaula `u0qj94fxxJ6` history `[1..383]`, `[385..485]`, `[487..531]`, `[533..1055]`, `[533..1055]`, `[1057..1590]`
   * this brought no further acceleration

brenzi:
new cli command `purge-community-ceremony` in order to resolve #191

### DIFF
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -5,9 +5,11 @@ const SIGNER_ARG: &'static str = "signer";
 const CID_ARG: &'static str = "cid";
 const CLAIMS_ARG: &'static str = "claims";
 const CEREMONY_INDEX_ARG: &'static str = "ceremony-index";
-const IPFS_CID_ARG: &'static str = "ceremony-index";
+const IPFS_CID_ARG: &'static str = "ipfs-cid";
 const BOOTSTRAPPER_ARG: &'static str = "bootstrapper";
 const FUNDEES_ARG: &'static str = "fundees";
+const FROM_CINDEX_ARG: &'static str = "from-cindex";
+const TO_CINDEX_ARG: &'static str = "to-cindex";
 const ENDORSEES_ARG: &'static str = "endorsees";
 const ALL_FLAG: &'static str = "all";
 
@@ -20,6 +22,8 @@ pub trait EncointerArgs<'b> {
 	fn ipfs_cid_arg(self) -> Self;
 	fn bootstrapper_arg(self) -> Self;
 	fn fundees_arg(self) -> Self;
+	fn from_cindex_arg(self) -> Self;
+	fn to_cindex_arg(self) -> Self;
 	fn endorsees_arg(self) -> Self;
 	fn all_flag(self) -> Self;
 }
@@ -33,6 +37,8 @@ pub trait EncointerArgsExtractor {
 	fn ipfs_cid_arg(&self) -> Option<&str>;
 	fn bootstrapper_arg(&self) -> Option<&str>;
 	fn fundees_arg(&self) -> Option<Vec<&str>>;
+	fn from_cindex_arg(&self) -> Option<i32>;
+	fn to_cindex_arg(&self) -> Option<i32>;
 	fn endorsees_arg(&self) -> Option<Vec<&str>>;
 	fn all_flag(&self) -> bool;
 }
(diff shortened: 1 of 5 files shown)


client/src/cli_args.rs
```diff
@@ -5,9 +5,11 @@ const SIGNER_ARG: &'static str = "signer";
 const CID_ARG: &'static str = "cid";
 const CLAIMS_ARG: &'static str = "claims";
 const CEREMONY_INDEX_ARG: &'static str = "ceremony-index";
-const IPFS_CID_ARG: &'static str = "ceremony-index";
```
clangenb, line 8: Haha, this kind of error will no longer be possible with clap v3

client/src/main.rs
```diff
@@ -1059,6 +1059,68 @@ fn main() {
                     Ok(())
                 }),
         )
+        .add_cmd(
+                Command::new("purge-community-ceremony")
+                    .description("purge all history within the provided ceremony index range for the specified community")
+                    .options(|app| {
```
clangenb, line 1065: Might want to add here: `if both arguments for the ceremony index range are omitted, the last ceremony index is purged.`

client/src/main.rs
```diff
@@ -1059,6 +1059,68 @@ fn main() {
...
+
+                    let cid = verify_cid(&api,
+                                         matches
+                                             .cid_arg()
+                                             .expect("please supply argument --cid"),
+                    );
+                    println!("purging ceremony index range [{}  {}] for community {}", from_cindex, to_cindex, cid);
+
+                    let calls: Vec<_> = (from_cindex..to_cindex+1)
+                        .map(|idx| compose_call!(
```
clangenb: I think this should be possible and more intuitive: `(from_cindex..=to_cindex)`

client/src/main.rs
```diff
@@ -1059,6 +1059,68 @@ fn main() {
...
+
+                    })
+                .runner(|_args: &str, matches: &ArgMatches<'_>| {
+                    let sudoer = AccountKeyring::Alice.pair();
+                    let api = get_chain_api(matches).set_signer(sudoer);
+
+                    let current_ceremony_index = get_ceremony_index(&api);
+
+                    let from_cindex_arg = matches.from_cindex_arg().unwrap_or(0);
+                    let to_cindex_arg = matches.to_cindex_arg().unwrap_or(0);
```
clangenb, line 1078: I guess we should assert here that `to_cindex_arg >from_index_arg` and give a meaningful error message.

clangenb (approved):
Some minor beauty-clarification details, but looks good otherwise.

Will preemptively approve, you can merge at your discretion.

client/src/main.rs
brenzi: no, the arguments are mandatory in this case ( so the transformation is actually not necessary....

client/src/main.rs
brenzi: true

client/src/main.rs
brenzi: true

PR #194, opened 2022-03-30
Fees in community currency

pifragile:
This is the node part of https://github.com/encointer/pallets/pull/190

this only solves the node part. the client doesn't support CC fees yet. (left for #195 to do)

EDIT CL:

For now, it depends on the hotfix-branch of the api-client to make the integration tests pass: https://github.com/scs/substrate-api-client/tree/cl/encointer-community-currency-fees-hotfix. The hotfix updates the signed-extension of the api-client to have a valid format for the newly introduced `ChargeAssetTxPayment`. Currently, it is hardcoded to always use the native token. Change that once we have implemented: https://github.com/scs/substrate-api-client/issues/211

Tested that community-fees work according to https://github.com/encointer/encointer-node/issues/200.

Todo before merge:
* Needs https://github.com/encointer/pallets/pull/195


brenzi:
this will further need to use signed extensions to define the fee paying asset. See Statemine for this
https://github.com/paritytech/cumulus/blob/71d7743869f765e1531a4e43ce278d03e4fd8808/polkadot-parachains/statemine/src/lib.rs#L586-L596

And we will need the client side for this too: 
https://github.com/scs/substrate-api-client/issues/211

node/src/chain_spec.rs
```diff
@@ -193,5 +193,7 @@ fn testnet_genesis(
 			min_solar_trip_time_s: 1, // [s]
 			max_speed_mps: 1,         // [m/s] suggested would be 83m/s,
 		},
+
+		encointer_balances: EncointerBalancesConfig { fee_conversion_factor: 10_000 },
```
brenzi: I'd like a comment here to explain what this value means and how to calculate it.

brenzi: like:
		// for relative adjustment.
			// 100_000 translates 5uKSM to 0.01 CC if ceremony reward is 20 CC
			// lower values lead to lower fees in CC proportionally

brenzi: ..and I therefore suggest to set the value to 100_000

pifragile: @clangenb i originally set it up such that a factor of 10_000 matches those constraints and I guess you changed that. What factor do we have to set here such that this constraint holds?

brenzi: your unit tests in pallet repo use 100_000 that's why I changed it

clangenb: Sorry, my fingers were in the game there. Yes `100_000` should it be now.

runtime/Cargo.toml
```diff
@@ -16,6 +16,7 @@ serde = { features = ["derive"], optional = true, version = "1.0.136" }
 
 # encointer deps
 encointer-primitives = { default-features = false, git = "https://github.com/encointer/pallets", branch = "master" }
+encointer-balances-tx-payment = { default-features = false, git = "https://github.com/encointer/pallets", branch = "master" }
```
clangenb, line 19: Always also include the `std` feature when adding dependencies with `default-features` false.

It might have unforeseen consequences when this is not done. Substrate for instance infers a wasm environment sometimes.

client/Cargo.toml
```diff
@@ -28,8 +28,8 @@ pallet-encointer-communities = { git = "https://github.com/encointer/pallets", b
 pallet-encointer-scheduler = { git = "https://github.com/encointer/pallets", branch = "master" }
 
 # scs deps
-substrate-api-client = { git = "https://github.com/scs/substrate-api-client", branch = "master"}
-substrate-client-keystore = { git = "https://github.com/scs/substrate-api-client", branch = "master"}
+substrate-api-client = { git = "https://github.com/scs/substrate-api-client", branch = "cl/encointer-community-currency-fees-hotfix"}
```
brenzi, line 31: not sure I'd like to merge that yet with this branch reference. maybe wait for a proper solution?

brenzi (changes requested):

brenzi (approved):

brenzi:
should be merged when CI passes

*edit: not possible to pass with this client based on api-client master. . will revert commit. I deem it acceptable to depend on a branch until #195 is fixed*

PR #196, opened 2022-04-03
[Client] refactor new_community CLI

clangenb:
It was not very readable before. The refactoring is a preparation for encointer/encointer-parachain/issues/109 in order to simplify the switch between sudo extrinsic and council proposal extrinsic.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -28,9 +28,9 @@ use crate::{
 		add_location_call, new_community_call, read_community_spec_from_file, CommunitySpec,
 	},
 	utils::{
-		batch_call, into_effective_cindex,
+		batch_call, ensure_payment, into_effective_cindex,
 		keys::{get_accountid_from_str, get_pair_from_str},
-		offline_xt, sudo_call, sudo_xt,
+		offline_xt, send_and_wait_for_in_block, sudo_call, sudo_xt,
 	},
 };
 use clap::{value_t, AppSettings, Arg, ArgMatches};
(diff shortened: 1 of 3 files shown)


client/src/main.rs
```diff
@@ -353,134 +359,40 @@ fn main() {
...
-                    ensure_payment(&api, &xt.hex_encode());
-                    let tx_hash = api.send_extrinsic(xt.hex_encode(), XtStatus::InBlock).unwrap();
-                    info!("[+] Transaction got included. Hash: {:?}\n", tx_hash);
-                    println!("{}", cid);
-
                     if api.get_current_phase().unwrap() != CeremonyPhaseType::REGISTERING {
-                        error!("wrong ceremony phase for registering new locations for {}", cid);
+                        error!("Wrong ceremony phase for registering new locations for {}", cid);
+                        error!("Exiting process without registering a new community");
                         std::process::exit(exit_code::WRONG_PHASE);
```
clangenb: Before this was called after registering the community. Was there a reason for that?

If not, I thought it is more intuitive to not register the community at all.

client/src/main.rs
```diff
@@ -1156,28 +1068,6 @@ fn reasonable_native_balance(api: &Api<sr25519::Pair, WsRpcClient>) -> u128 {
 	return ed + fee * PREFUNDING_NR_OF_TRANSFER_EXTRINSICS
 }
 
-fn ensure_payment(api: &Api<sr25519::Pair, WsRpcClient>, xt: &str) {
```
clangenb, line 1159: moved as is to `utils`

client/src/main.rs
brenzi: Registering a new community is always allowed. Registering new locations only during REGISTERING phase. I'd leave it like that because it's handy. Also, we may still want to call this function just to dump the encoded call data to submit with different origin via js/apps

clangenb: Ok, I changed that back to how it was before.

brenzi (approved):
Great improvement for readability!

PR #197, opened 2022-04-03
Enable sending privileged extrinsics with a council-proposal

clangenb:
When the `sudo` pallet does not exist, we infer that we are talking to a chain with the `collective` pallet where `Alice` is the sole member. Privileged calls are sent as a council proposal as `Alice` leading to immediate execution of the proposal if successful.

Tested against encointer-parachain with Alice as sole council member:
* `new-community` and `next-phase` CLI work
* `bootstrap_demo_community.py` runs successfully

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -28,7 +28,7 @@ use crate::{
 		add_location_call, new_community_call, read_community_spec_from_file, CommunitySpec,
 	},
 	utils::{
-		batch_call, ensure_payment, into_effective_cindex,
+		batch_call, contains_sudo_pallet, ensure_payment, into_effective_cindex,
 		keys::{get_accountid_from_str, get_pair_from_str},
 		offline_xt, send_and_wait_for_in_block, sudo_call, sudo_xt,
 	},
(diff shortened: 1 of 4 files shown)


clangenb:
@brenzi needs cleanup but you might want to comment already, as I need to go out now.

client/src/main.rs
```diff
@@ -374,26 +375,41 @@ fn main() {
...
+                        if api.get_current_phase().unwrap() != CeremonyPhaseType::REGISTERING {
+                            error!("Wrong ceremony phase for registering new locations for {}", cid);
+                            error!("Aborting without registering additional locations");
+                            std::process::exit(exit_code::WRONG_PHASE);
+                        }
+                        send_and_wait_for_in_block(&api, sudo_xt(&api, add_location_batch_call));
 
-                    send_and_wait_for_in_block(&api, sudo_xt(&api, add_location_batch_call));
+                    } else {
+                        info!("Printing raw collective propose calls for js/apps for cid: {}", cid);
```
brenzi, line 389: Quite some redundant code here. I think we could do the if only around the call wrapping and leave the rest the same for the two cases

brenzi: ;-)

brenzi:
Looks promising

brenzi:
Will you do next_phase at the same go, or a different pr?

client/src/main.rs
clangenb: If I understand you correctly. It is not as simple as you might think. In rust you can't do:

```
let call = if contains_sudo_pallet(&api.metadata) {
                        sudo_call(sudo_call(&api.metadata, new_community_call.clone())
                    } else {
                        collective_propose_call(&api.metadata, 1, new_community_call)
};
```

Because they are different types.

clangenb: Solved by introducing `OpaqueCall` type.

brenzi (approved):

PR #201, opened 2022-04-05
implement aggregated account data rpc

pifragile:


clangenb (approved):
LGTM, did you actually test that the rpc returns the correct value? An example of the returned result would be cool.

brenzi:
what about the cli client? don't we need to adapt namings there?

...we need to: please scan the cöli client for rpc calls and rename there. it seems none of them are tested in CI as everything passed, but that's just our bad coverage

i.e.:
https://github.com/encointer/encointer-node/blob/dca45027de7dd60f6a49282efca2dd772462c537/client/src/main.rs#L1341-L1347

clangenb:
This does only add an RPC, which has not been integrated in the CLI yet, but the RPCs from before they still work. I suggest doing this in a separate task. So we have it at least in the runtime, I'd just like to test it in polkadot-js/apps to see that it works.

brenzi:
haha, ok, I was mislead. pardon. go ahead!

clangenb:
I just tested that the rpc exists in the retrieved by polkadot-js/apps

PR #204, opened 2022-04-12
change RPC prefixes

pifragile:
Change RPC prefixes according to https://github.com/encointer/pallets/commit/dc863acf408e552dd7177881824c895856f1f516

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1392,7 +1392,7 @@ dependencies = [
 [[package]]
 name = "encointer-balances-tx-payment"
 version = "1.0.0"
-source = "git+https://github.com/encointer/pallets?branch=master#bbd904649152c6db763c7bb476b5bab0f89edb79"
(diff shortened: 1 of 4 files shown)


brenzi:
I guess you didn't downgrade substrate. You just reverted the accidental upgrade cargo did when bumping pallets, right?

pifragile:
> I guess you didn't downgrade substrate. You just reverted the accidental upgrade cargo did when bumping pallets, right?

yes😅

clangenb (approved):
CI passes so ready to merge from my side.

I made an issue for updating the prefixes in https://github.com/encointer/encointer-js/issues/59

clangenb:
@brenzi Do you want to wait, or can we merge this?

brenzi:
We missed the window of opportunity to get it into the parachain runtime. If we can bend it to work with our para runtime with only updating the collator binary, fine, then we should move forward asap and align all interfaces including the mobile app and ci

clangenb:
This does not touch the runtime, we should be able to update the collator without any problems.

PR #226, opened 2022-06-06
Allow paying fees in CC and add integration test for transfer_all

### LINKED ISSUE #195: cli client can select to pay fees in community currency
Upgrade CI directly to test this in e2e

* [ ] bump api-client to get: https://github.com/scs/substrate-api-client/commit/ab586d8be719243aa29503ec8c7b1457b2cd0830 and https://github.com/scs/substrate-api-client/commit/edf0c13b05e43d9550a0a3773915c7bee74a48b5
* [ ] Define our custom signed extension where the [AssetTip.asset](https://github.com/scs/substrate-api-client/blob/edf0c13b05e43d9550a0a3773915c7bee74a48b5/primitives/src/extrinsic_params.rs#L225) is of type `Option<CommunityIdentifer>`
* [ ] re-enable `ensure_payment` wherever it has been commented. Change its signature to have `Option<CommunityIdentifier>`. If the cid is not none, simply check that the supplied account has some CC.
* [ ] Add `--payment-cid` argument for all commands that send an extrinsic. If it is not passed, we infer the native currency.
* [ ] add a section at the end of `bootstrap_demo_community.py` that tries to send `transfer_all` and pay in CC after the community income was distributed.

PoC: on how to pay in assets: https://github.com/scs/substrate-api-client/commit/b51d802b054bbd655b3178448822e2db04a8a71e

pifragile:
* Closes #195


client/src/utils.rs
```diff
@@ -92,26 +92,46 @@ pub fn contains_sudo_pallet(metadata: &Metadata) -> bool {
...
+				&api.signer_account().unwrap(),
+				None,
+			)
+			.unwrap()
+		{
+			None => {
+				error!("insufficient balance in community {}", cid);
+				std::process::exit(exit_code::FEE_PAYMENT_FAILED);
+			},
+			_ => (),
```
clangenb: It would be super useful to see how much CC the signer has.

Also, please separate this in two function to make it more readable:

```
pub fn ensure_payment(...) {
    match tx_payment_cid {
       Some(cid) => ensure_cc_payment(...),
       None => ensure_native_payment(..) 
   }
}
```

client/src/main.rs
```diff
@@ -1557,7 +1639,7 @@ fn endorse_newcomers(
 
 		let xt = offline_xt(&api, call, nonce);
 
-		ensure_payment(&api, &xt.hex_encode());
+		ensure_payment(&api, &xt.hex_encode(), None);
```
clangenb: Why did you put `None` here?

client/src/main.rs
```diff
@@ -1519,18 +1599,20 @@ fn apply_demurrage(
 fn send_bazaar_xt(matches: &ArgMatches<'_>, business_call: &BazaarCalls) -> Result<(), ()> {
 	let business_owner = matches.account_arg().map(get_pair_from_str).unwrap();
 
-	let api = get_chain_api(matches).set_signer(business_owner.clone().into());
+	let mut api = get_chain_api(matches).set_signer(business_owner.clone().into());
 	let cid = verify_cid(&api, matches.cid_arg().expect("please supply argument --cid"));
 	let ipfs_cid = matches.ipfs_cid_arg().expect("ipfs cid needed");
 
+	let tx_payment_cid_arg = matches.tx_payment_cid_arg();
+	api = set_api_extrisic_params_builder(api.clone().into(), tx_payment_cid_arg);
```
clangenb: Here, performance does not really matter, but the clone is still not necessary. :)

clangenb: The same is true for all the other cases.

client/py_client/client.py
```diff
@@ -217,3 +236,9 @@ def get_bootstrappers_with_remaining_newbie_tickets(self, cid):
         ret = subprocess.run(self.cli + ["get-bootstrappers-with-remaining-newbie-tickets", "--cid", cid],
                              stdout=subprocess.PIPE)
         return ret.stdout.decode("utf-8").strip()
+
+    def transfer_all(self, cid, source, dest, pay_fees_in_cc=False):
```
clangenb, line 223: Looks very good, but can you add the `pay_fees_in_cc` for all commands that send an extrinsic here. We will use that in the future for better integration tests.

If I saw correctly, you already check this CLI on the rust side already anyhow. :)

clangenb (commented):
Looks very nice, only some minor comments.

client/src/utils.rs
pifragile: well, 0 balance :) i changed the message

clangenb: ah sorry, the `_ => ()` irritated my eyes, I thought you just ignore that. :)

All good then!

client/py_client/client.py
pifragile: i ended up refactoring the whole thing, see 44ad2d9845e89fc0d2cf29acf2db2b0a8fb5c807

clangenb: Looks very good! Thanks!

client/src/utils.rs
```diff
@@ -65,7 +65,7 @@ pub fn send_and_wait_for_in_block<C: Encode>(api: &Api, xt: EncointerXt<C>) -> O
 }
 
 pub fn send_xt_hex_and_wait_for_in_block(api: &Api, xt_hex: String) -> Option<H256> {
-	ensure_payment(&api, &xt_hex);
+	ensure_payment(&api, &xt_hex, None);
```
clangenb: I know it is a bit tedious, but I would like to have the payment_cid arg passed here too.

clangenb (commented):
Looks very good, only one minor things remains!

Cargo.lock
```diff
@@ -7061,7 +7061,7 @@ dependencies = [
 [[package]]
 name = "sp-arithmetic"
 version = "5.0.0"
-source = "git+https://github.com/paritytech/substrate.git?branch=master#699375ef8dd0cb19f0b5d8917e4e2a02d50c92bf"
+source = "git+https://github.com/paritytech/substrate.git?branch=master#fe5bf49290d166b9552f65e751d46ec592173ebd"
```
clangenb: Please downgrade the substrate deps again:

```
cargo update -p sp-io --precise 699375ef8dd0cb19f0b5d8917e4e2a02d50c92bf
```

clangenb (commented):
Looks good to me except for the minor comment, and we need to fix the bazaar integration test, see:

https://github.com/encointer/encointer-node/runs/6814147967?check_suite_focus=true

clangenb (approved):
Looks good to me!

brenzi:
@pifragile you seem to have changed formatting settings in python. this makes it harder to review. It seems your line length is 80 whereas we usually use 120

brenzi (changes requested):
worked for me.
please revert the unnecessary python formatting changes, ripping apart the print statements

I think our final transfer_all test could be more thorough:
1. `transfer` exactly 0.5 CC from `//Alice` to `//newbie1`, an account which provably has no native balance to pay fees with. This transfer could be explicitly set to pay fees in ERT, to test that too
2. `transfer_all` from `//newbie1` to `//newbie2`
3. ensure that `//newbie1` has zero balance
4. ensure that `//newbie2` has nonzero balance

this way, we also test that fee payment in CC works

brenzi (approved):

PR #227, opened 2022-06-11
Cli add locations

brenzi:


client/encointer-api-client-extension/src/ceremonies.rs
```diff
@@ -326,7 +326,7 @@ impl CeremoniesApi for Api {
 		// get stats of every meetup
 		for m in 1..=mcount {
 			let m_location = self.get_meetup_location(&community_ceremony, m)?.unwrap();
-			let time = self.get_meetup_time(m_location, ONE_DAY)?;
+			let time = self.get_meetup_time(m_location, ONE_DAY).unwrap_or(0);
```
clangenb: What is the motivation here?

client/src/main.rs
```diff
@@ -477,6 +477,64 @@ fn main() {
...
+                    let tx_payment_cid_arg = matches.tx_payment_cid_arg();
+
+                    let cid = verify_cid(&api, matches.cid_arg().unwrap(), None);
+
+                    let add_location_calls = spec.locations().into_iter().map(|l|
+                                                                                  {
+                                                                                      info!("adding location {:?}", l);
+                                                                                      add_location_call(&api.metadata, cid, l)
+                                                                                  }
+                        ).collect();
```
clangenb, line 512: fmt: :)

clangenb (approved):
I have a question and a remark, but looks good to me otherwise.

client/encointer-api-client-extension/src/ceremonies.rs
brenzi: needed that to make sth work, but will remove as it seems not necessary now

PR #229, opened 2022-06-13
Voucher Format update

clangenb:
* Add documentation
* Update voucher generation

The app successfully scanned this Code generated by: `python voucher.py --cid u0qj92QX9PQ --issuer "Verein Encointer" -n 10`

voucher-j9bSDKVnjVh-6FdFqMw7Tevvt8B5nGC7jfN9tfQt8BETx

Edit: Updated QR-code Image with v2 QR-code.

### DIFF
--- a/client/voucher.py
+++ b/client/voucher.py
@@ -23,7 +23,7 @@ def main(issuer, cid, network, n):
             voucher_uri = f"//{batch_token}/{voucher_token}"
             f.write(voucher_uri)
             print(f"generating voucher {i}: {voucher_uri}")
-            payload = f"encointer-voucher\nV1.0\n{voucher_uri}\n{cid}@{network}\n\n{issuer}"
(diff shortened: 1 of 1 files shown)


client/voucher.py
```diff
@@ -23,7 +23,7 @@ def main(issuer, cid, network, n):
             voucher_uri = f"//{batch_token}/{voucher_token}"
             f.write(voucher_uri)
             print(f"generating voucher {i}: {voucher_uri}")
-            payload = f"encointer-voucher\nV1.0\n{voucher_uri}\n{cid}@{network}\n\n{issuer}"
+            payload = f"encointer-voucher\nV1.0\n{voucher_uri}\n{cid}\n{network}\n{issuer}"
```
brenzi: shouldn't the version read 2.0 now?

clangenb: Ah yes, but as the voucher does only exist in v2.0, it is the only one that does not check the version. If it can decode the field into the `enum` with values v1.0 and v2.0 it treats it as a success.

brenzi (approved):

PR #232, opened 2022-06-20
Bump substrate

pifragile:


Cargo.lock
```diff
@@ -1874,7 +1886,7 @@ checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 [[package]]
 name = "fork-tree"
 version = "3.0.0"
-source = "git+https://github.com/paritytech/substrate.git?branch=master#699375ef8dd0cb19f0b5d8917e4e2a02d50c92bf"
```
clangenb, line 1877: Something seems to have gone wrong in the update, there are still quite some `699375ef8dd0cb19f0b5d8917e4e2a02d50c92bf` references in the lock file.

Why did I notice this? It seemed very weird that there were 505 additions and only 272 deletions in the lock file.

Cargo.lock
```diff
@@ -2803,25 +2826,39 @@ version = "0.13.1"
...
- "jsonrpsee-core",
- "jsonrpsee-http-server",
- "jsonrpsee-proc-macros",
- "jsonrpsee-types",
+ "jsonrpsee-core 0.13.1",
+ "jsonrpsee-http-server 0.13.1",
+ "jsonrpsee-proc-macros 0.13.1",
+ "jsonrpsee-types 0.13.1",
+ "jsonrpsee-ws-server 0.13.1",
+ "tracing",
```
clangenb, line 2834: This is also an indicator that something might have gone wrong. Cargo add the versions of the dependencies when we have the same deps twice with different versions.

node/Cargo.toml
```diff
@@ -20,66 +20,65 @@ targets = ["x86_64-unknown-linux-gnu"]
...
-sc-consensus = { version = "0.10.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sc-finality-grandpa = { version = "0.10.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sp-finality-grandpa = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sc-client-api = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sp-timestamp = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sp-inherents = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-sp-keyring = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git",branch = "master" }
-frame-system = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git",branch = "master" }
+frame-system = {version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "master"}
```
clangenb, line 27: Here we should not have autoformating enabled because upstream does not use it. This makes upstream merges harder

clangenb:
There is a new one incoming

PR #235, opened 2022-06-24
[cli] At block arg for balance, add signer arg for next-phase

brenzi:
new cli arguments
* `--at` to specify block hash at which to query balance 
* `--signer` to specify extrinsic signer by SS58, suri, seed or mnemonic

also, a helper script has been added for reference to dump balances for a list of accounts

the Leu config changes are a bit alien to this PR but they are related to the need for the new arguments. so leaving that commit in here

usage to query historic balance
```
../target/release/encointer-client-notee -u wss://kusama.api.encointer.org -p 443 balance 5Cm38saAbnwrxvnGCp4Pqff5K3rKy4awYQnrdqDeHfvW2ZUM --cid u0qj92QX9PQ --at 0xcfb07c60aadd57676ce0591678b58511ebd03bdef7385c9690f42e744f1dbff6
```

### DIFF
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -69,9 +69,11 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
 	fn signer_arg(self, help: &'b str) -> Self {
 		self.arg(
 			Arg::with_name(SIGNER_ARG)
+				.short("s")
+				.long("signer")
 				.takes_value(true)
-				.required(true)
-				.value_name("SS58")
+				.required(false)
+				.value_name("suri, seed , mnemonic or SS58 in keystore")
 				.help(help),
 		)
 	}
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -1,4 +1,6 @@
 use clap::{App, Arg, ArgMatches};
+use substrate_api_client::Hash;
+use substrate_api_client::FromHexString;
 
 const ACCOUNT_ARG: &'static str = "accountid";
 const SIGNER_ARG: &'static str = "signer";
@@ -14,6 +16,8 @@ const ENDORSEES_ARG: &'static str = "endorsees";
 const TIME_OFFSET_ARG: &'static str = "time-offset";
 const ALL_FLAG: &'static str = "all";
 const TX_PAYMENT_CID_ARG: &'static str = "tx-payment-cid";
+const AT_BLOCK_ARG: &'static str = "at";
+
 
 pub trait EncointerArgs<'b> {
 	fn account_arg(self) -> Self;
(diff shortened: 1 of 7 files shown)


client/src/cli_args.rs
```diff
@@ -251,4 +268,11 @@ impl<'a> EncointerArgsExtractor for ArgMatches<'a> {
 	fn tx_payment_cid_arg(&self) -> Option<&str> {
 		self.value_of(TX_PAYMENT_CID_ARG)
 	}
+	fn at_block_arg(&self) -> Option<Hash> {
+		if let Some(hexhashstr) = self.value_of(AT_BLOCK_ARG) {
+			Some(Hash::from_hex(hexhashstr.to_string()).unwrap())
+		} else {
+			None
+		}
+	}
```
clangenb: This is the verbose version of:

```
self.value_of(AT_BLOCK_ARG)
     .map(|hex| Hash::from(hex.to_string().unwrap())
```

client/src/main.rs
```diff
@@ -492,12 +494,18 @@ fn main() {
...
+                        .at_block_arg()
+                })
                 .runner(|_args: &str, matches: &ArgMatches<'_>| {
                     let api = get_chain_api(matches);
+                    let maybe_at = matches.at_block_arg();
                     let cid = verify_cid(&api,
                                          matches
                                              .cid_arg()
                                              .expect("please supply argument --cid"),
+                        maybe_at
```
clangenb: Ui, we should enable rustfmt at some point here.

client/src/main.rs
```diff
@@ -530,9 +538,18 @@ fn main() {
...
+                .options(|app| {
+                    app.setting(AppSettings::ColoredHelp)
+                        .signer_arg("account with necessary privileges (sudo or councillor)")
+                })
                 .runner(|_args: &str, matches: &ArgMatches<'_>| {
+                    let signer = if let Some(sig) = matches.signer_arg() {
+                        get_pair_from_str(sig).into()
+                    } else {
+                        AccountKeyring::Alice.pair()
+                    };
```
clangenb: This is also a verbose version of:

```
let signer = matches.signer_arg()
    .map_or_else(|| Alice.pair(), |signer| get_pair_from_str(signer).into());
```

client/src/main.rs
```diff
@@ -953,7 +972,11 @@ fn main() {
 
                     extract_and_execute(
                         &matches, |api, cid| {
-                            let signer = matches.signer_arg().map(get_pair_from_str).unwrap();
+                            let signer = if let Some(sig) = matches.signer_arg() {
+                                get_pair_from_str(sig)
+                            } else {
+                                panic!("please specify --signer. must be an assignee of a recent meetup")
+                            };
```
clangenb: I generally think match statements are more readable if both branches of an `if let Some(..)` statement are needed.

clangenb (commented):
Only some minor style comments, looks good otherwise.

client/src/cli_args.rs
brenzi: haha, thought you might not like these ;-)

PR #238, opened 2022-07-25
bump pallets and add integration tests

pifragile:
This PR includes the integration tests for the RPC caching and fee refunding which comes with a refactor of the `bootstrap_demo_community` script.


brenzi (approved):
if this PR includes a refactoring of the py client, the OP should explain this

otherwise, fine for me

PR #243, opened 2022-07-27
Adjust CLI to runtime v8

### LINKED ISSUE #241: [cli] allow claim-reward by anyone
since https://github.com/encointer/pallets/pull/244
anyone can claim on behalf of any meetup.

the cli client should support this feature. the subcommand "claim-reward" should take an optional argument being the meetup index to be claimed, or, additionally `--all` which sends a batch extrinsic to claim all meetups up to the assigned number

### LINKED ISSUE #242: [cli] allow registering in attesting phase
currently, the cli prevents this outside of registering phase, but since https://github.com/encointer/pallets/pull/224 we may register earlier

please adjust the check and also include the `--signer` argument to optionally pass the mnemonic instead of using the keystore

pifragile:
* Closes #241
* Closes #242 

introduces a `--meetup-index` and `--all` flag for `claim-rewards`
and allows registering in attesting phase
and adds optional `--signer` arg to `register-participant`

### DIFF
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -16,7 +16,6 @@ const TIME_OFFSET_ARG: &'static str = "time-offset";
 const ALL_FLAG: &'static str = "all";
 const TX_PAYMENT_CID_ARG: &'static str = "tx-payment-cid";
 const MEETUP_INDEX_ARG: &'static str = "meetup-index";
-const ALL_UPTO_ARG: &'static str = "all-upto";
 const AT_BLOCK_ARG: &'static str = "at";
 
 pub trait EncointerArgs<'b> {
@@ -35,7 +34,6 @@ pub trait EncointerArgs<'b> {
 	fn all_flag(self) -> Self;
 	fn tx_payment_cid_arg(self) -> Self;
 	fn meetup_index_arg(self) -> Self;
-	fn all_upto_arg(self) -> Self;
 	fn at_block_arg(self) -> Self;
 }
 
@@ -55,7 +53,6 @@ pub trait EncointerArgsExtractor {
 	fn all_flag(&self) -> bool;
 	fn tx_payment_cid_arg(&self) -> Option<&str>;
 	fn meetup_index_arg(&self) -> Option<u64>;
-	fn all_upto_arg(&self) -> Option<u64>;
 	fn at_block_arg(&self) -> Option<Hash>;
 }
 
@@ -216,18 +213,8 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
 				.takes_value(true)
 				.required(false)
 				.value_name("MEETUP_INDEX")
+				.conflicts_with(ALL_FLAG)
 				.help("the meetup index for which to claim rewards"),
-		)
-	}
-
-	fn all_upto_arg(self) -> Self {
-		self.arg(
-			Arg::with_name(ALL_UPTO_ARG)
-				.long("all-upto")
-				.takes_value(true)
-				.required(false)
-				.value_name("ALL_UPTO")
-				.help("claim rewards for meetup indexes 1..=all-upto"),
 		)
 	}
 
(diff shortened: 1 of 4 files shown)


client/src/cli_args.rs
```diff
@@ -32,6 +34,8 @@ pub trait EncointerArgs<'b> {
 	fn time_offset_arg(self) -> Self;
 	fn all_flag(self) -> Self;
 	fn tx_payment_cid_arg(self) -> Self;
+	fn meetup_index_arg(self) -> Self;
+	fn all_upto_arg(self) -> Self;
```
brenzi: I can live with this, but why not just --all? The cli can find out the number of assigned meetups itself

client/src/cli_args.rs
```diff
@@ -202,6 +208,28 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
...
+				.takes_value(true)
+				.required(false)
+				.value_name("MEETUP_INDEX")
+				.help("the meetup index for which to claim rewards"),
+		)
+	}
+
+	fn all_upto_arg(self) -> Self {
+		self.arg(
+			Arg::with_name(MEETUP_INDEX_ARG)
```
brenzi: copy-paste error. guess you didn't test this? 🤭

client/src/cli_args.rs
```diff
@@ -202,6 +208,28 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
...
+				.long("all-upto")
+				.takes_value(true)
+				.required(false)
```
brenzi: I'd like to see `.conflicts_with("meetup-index")`. It doesn't make sense to specify both

client/src/cli_args.rs
```diff
@@ -202,6 +208,28 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
...
+				.value_name("MEETUP_INDEX")
```
brenzi: also here, copy-paste error

client/src/cli_args.rs
brenzi: hmm, I really would prefer this to be `--all`

client/src/main.rs
```diff
@@ -1034,31 +1040,62 @@ fn main() {
...
-                                ENCOINTER_CEREMONIES,
-                                "claim_rewards",
-                                cid,
-                                Option::<MeetupIndexType>::None
-                            );
-                            ensure_payment(&api, &xt.hex_encode(), tx_payment_cid_arg);
 
-                            let _ = api.send_extrinsic(xt.hex_encode(), XtStatus::Ready).unwrap();
-                            println!("Claiming reward for {}. xt-status: 'ready'", signer.public());
+                            if let Some(all_upto) = all_upto_arg {
```
brenzi: now you do nothing if neither meetup-index, nor all_upto were specified. That is not what I wanted. Please leave the old behaviour if no extra args are specified

brenzi (changes requested):

client/src/cli_args.rs
pifragile: i thought that this would become a problem😅 the all flag is already taken and takes no argument this is why i did it like this. is there an easy solution to this?

client/src/cli_args.rs
pifragile: oh good catch. i actually did test an it worked, but just with one meetup, maybe this is why is still worked. or could it maybe have worked because the `.long` is set correclty?

client/src/cli_args.rs
pifragile: oh perfect i didn't know that this existed

client/src/main.rs
pifragile: well this is what i am doing, setting `meetup_index` to `None` which leads to the old behaviour

client/src/cli_args.rs
pifragile: oh yes that would be possible. i misread the issue and thought that you wanted to be able to specify an upper bound

brenzi (approved):

PR #247, opened 2022-08-10
benchmark upgrade_registration and unregister

pifragile:
Benchmarks for https://github.com/encointer/pallets/pull/254


brenzi:
I think we need a different process for benchmarks than such PR's. same machine always, alerts upon significant weight changes....

I now just copy-pasted weights from pallets into #248. not the clean approach as its also not the right PR, but if we want to keep node and pallet in different repositories, it's too much hassle to review and test every PR with its counterpart on this repo

brenzi:
I'll close it as #248 includes benchmarks

PR #248, opened 2022-08-10
add TrustableForNonDestructiveAction

pifragile:
node side of https://github.com/encointer/pallets/pull/255

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -61,7 +61,7 @@ pub use encointer_primitives::{
 	communities::{CommunityIdentifier, Location},
 	scheduler::CeremonyPhaseType,
 };
-use frame_system::EnsureRoot;
+use frame_system::{EnsureRoot, EnsureSigned};
 
 mod weights;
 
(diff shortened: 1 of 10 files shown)


runtime/src/lib.rs
```diff
@@ -440,6 +440,7 @@ impl pallet_encointer_ceremonies::Config for Runtime {
 impl pallet_encointer_communities::Config for Runtime {
 	type Event = Event;
 	type CommunityMaster = EnsureRoot<AccountId>;
+	type TrustableForNonDestructiveAction = EnsureRoot<AccountId>;
```
brenzi: This is meant to be 'EnsureSigned' AKA 'anyone who pays fees' because in the case of solochains (Gesell) we want to be permissive

brenzi (changes requested):

runtime/src/lib.rs
brenzi: Please test in CI e2e that anyone can register a community

pifragile: oh ok, i wasn't sure about that and i wanted to be conservative with security :) does that generally mean that everything in `node/runtime` will only ever be run on testnets?

brenzi: Currently yes

clangenb: This is true only for the node though. The parachain is productive by default.

brenzi:
@pifragile please ensure that your branch is up to date when you request my review. I first want to see CI passing respective to master

brenzi (approved):
let's merge this. I'll want to change the behaviour of cli and bot-community scripts, but not in this PR

PR #249, opened 2022-08-14
Second Leu rescue tooling

brenzi:
Helper scripts for rescue actions, useful for various purposes

todo
* extend dryrun flag to all extrinsics -> #250


clangenb (approved):
Looks good to me, but the client directory is getting too loaded for my taste. We should re-organize it soon.

PR #253, opened 2022-08-27
cli Ux improvements

### LINKED ISSUE #255: [cli] allow querying ceremony history during any ceremony phase
currently, querying ceremony stats for past ceremonies fails during registering phase:
```
listing meetups for cid u0qj944rhWE and ceremony nr 15
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Other("ceremony phase must be Assigning or Attesting to request meetup location.")', client/src/main.rs:734:102
```

this error doesn't make sense for historical data and should be fixed

brenzi:
closes #255 

* better error msg when account not in keystore
* dryrun flag for add-locations
* ceremony stats should not panic for past ceremony index (timestamp can't be reproduced!)

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -501,6 +501,7 @@ fn main() {
                 .options(|app| {
                     app.setting(AppSettings::ColoredHelp)
                         .signer_arg("account with necessary privileges")
+                        .dryrun_flag()
                         .arg(
                             Arg::with_name("specfile")
                                 .takes_value(true)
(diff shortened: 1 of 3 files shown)


client/src/main.rs
```diff
@@ -546,13 +550,18 @@ fn main() {
...
                     }
-                    // ---- send xt's to chain
-                    if api.get_current_phase().unwrap() != CeremonyPhaseType::Registering {
-                        error!("Wrong ceremony phase for registering new locations for {}", cid);
-                        error!("Aborting without registering additional locations");
-                        std::process::exit(exit_code::WRONG_PHASE);
+
+                    if matches.dryrun_flag() {
+                        println!("0x{}", hex::encode(add_location_batch_call.encode()));
+                    } else {
```
clangenb, line 556: From what I can see in the diff, it does not make sense to check for the dry_run flag twice. I'd prefer that the branching occurs only once.

clangenb (commented):
Once comment, but looks good otherwise

client/src/main.rs
brenzi: As discussed, I'll leave it because we need a more fundamental cli refactoring to solve this nicely

PR #256, opened 2022-09-07
fetch helper logs 'Issued' events too

brenzi:
Since runtime v8, ceremony reward payout also emits an Event per rewarded account. This should be tracked to yield correct balances


clangenb (approved):
Nice, looks good!

clangenb (approved):

PR #269, opened 2022-11-22
Integration tests for new features

pifragile:
Integration tests for: `upgrade_registration`, `unregister_participant`, endorsement after registration and endorsements by reputables.

Also there is a new flag in `bootstrap_demo_community.py`:
`bootstrap_demo_community.py` just bootstraps the community
`bootstrap_demo_community.py --test` runs the integration tests

### DIFF
--- a/client/bootstrap_demo_community.py
+++ b/client/bootstrap_demo_community.py
@@ -222,14 +222,47 @@ def test_unregister_and_upgrade_registration(client, cid):
     check_reputation(client, cid, newbie, 6, "VerifiedUnlinked")
 
 
+def test_endorsements_by_reputables(client, cid):
+    newbies = client.create_accounts(7)
+    faucet(client, cid, newbies)
 
+    register_participants_and_perform_meetup(client, cid, accounts + newbies[:1])
+    client.next_phase()
+    client.await_block(1)
+    claim_rewards(client, cid, account1, pay_fees_in_cc=True)
+    client.await_block(1)
+    # newbies[0] is now reputable
+    check_participant_count(client, cid, "Endorsee", 0)
+
+    # endorsement works before registration
+    client.endorse_newcomers(cid, newbies[0], [newbies[1]])
+    client.await_block(1)
+    client.register_participant(newbies[1], cid)
+    client.await_block(1)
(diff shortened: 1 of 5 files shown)


client/bootstrap_demo_community.py
```diff
@@ -28,13 +28,29 @@
...
 
-def perform_meetup(client, cid):
+def check_participant_count(client, cid, type, number):
+    participants_list = client.list_participants(cid)
+    print(participants_list)
+    expected_string = f"""Querying {type}Registry
+number of participants assigned:  {number}"""
+    if not expected_string in participants_list:
+        print(f"ERROR: Not {number} {type}s registered")
+        exit(1)
```
clangenb, line 38: Nice, I like this check.

client/bootstrap_demo_community.py
```diff
@@ -168,31 +184,109 @@ def test_reputation_caching(client, cid, account):
...
+
     if(not balance == client.balance(account1)):
         print("claim_reward fees were not refunded if paid in native currency")
         exit(1)
 
     client.next_phase()
     client.await_block(1)
 
+    if(not test):
+        print(f"Community {cid} sucessfullly bootatrapped")
```
clangenb, line 287: ah nice. Simple and easy. :)

client/src/main.rs
```diff
@@ -954,6 +954,125 @@ fn main() {
...
+                        None
+                    );
+
+
+                    let  mut cc: Option<CommunityCeremony> = None;
+                    if let Some(cindex_arg) = matches.ceremony_index_arg() {
+                        let current_ceremony_index = get_ceremony_index(&api);
+                        let cindex = into_effective_cindex(cindex_arg, current_ceremony_index);
+                        cc = Some((cid, cindex));
+                    }
```
clangenb: Minor thing, I would prefer:

```rust
let cc = match matches.ceremony_index_arg() {
   Some(cindex_arg) => ...,
   None => None,
}
```

because then it is obvious that `cc` is not mutated afterwards.

clangenb (commented):
Nice, it looks very clean! I only have minor remarks. 👍

clangenb (approved):
Perfect, thanks for the quick action!

PR #277, opened 2022-11-26
Publish test package

clangenb:
I want to easily download all the necessities to create demo communities in another CI. So I thought it would be usefull if we upload all the necessities as an artefact.


clangenb:
Superseded by the docker image

PR #281, opened 2022-12-02
fix cid decoding

### LINKED ISSUE #273: fetch-account-history.py broken since Kigali Aslah was registered
```
>./fetch-account-history.py <apikey>

scanning block 1689688 (44.32497013142174%) page : 0 - nr of events: 1, eta 0.14980362937617495h
Traceback (most recent call last):
  File "./fetch-account-history.py", line 85, in <module>
    cid = cid_raw['geohash'] + b58encode(bytearray.fromhex(cid_raw['digest'][2:])).decode("utf-8")
ValueError: non-hexadecimal number found in fromhex() arg at position 0
```
The reason is https://github.com/encointer/encointer-parachain/issues/155

The fix should be:  Make decoding conditional based on `0x` prefix here:
https://github.com/encointer/encointer-node/blob/c1c4b6571ca2a4d23561db8b4520fa147f2f6346/client/fetch-account-history.py#L79

brenzi:
closes #273

### DIFF
--- a/client/fetch-account-history.py
+++ b/client/fetch-account-history.py
@@ -19,7 +19,7 @@ page_rows = 100
 t_start = time.time()
 
 # TODO: edit these numbers to what you need. should be turned into cli arguments with defaults to 0..HEAD
-start_block = 784682# 1689688 # 1187284 #781981 #535620
+start_block = 1689688 #784682# # 1187284 #781981 #535620
 end_block = 1690155 #660815
 blocks_total = end_block - start_block
 
@@ -29,7 +29,7 @@ def decode_cid(cid_raw):
     if digest_raw.startswith('0x'):
         digest = b58encode(bytearray.fromhex(digest_raw[2:])).decode("utf-8")
     else:
-        digest = digest_raw
+        digest = b58encode(digest_raw).decode("utf-8")
     return geohash + digest
 
 
--- a/client/fetch-account-history.py
+++ b/client/fetch-account-history.py
@@ -18,10 +18,19 @@ page_rows = 100
 
 t_start = time.time()
 
-# TODO: edit these numbers to what you need. should be turned into cli arguments with defaults to 0..HEAD
-start_block = 1111286
-end_block = 1204517
+start_block = 1689688 # 1187284 #781981 #535620
+end_block = 1690155 #660815
 blocks_total = end_block - start_block
+
+def decode_cid(cid_raw):
+    geohash = cid_raw['geohash']
+    digest_raw = cid_raw['digest']
+    if digest_raw[0:1] == '0x':
+        digest = b58encode(bytearray.fromhex(digest_raw[2:])).decode("utf-8")
+    else:
+        digest = digest_raw
+    return geohash + digest
+
 
 with open(f'account-events-{account}-{start_block}_to_{end_block}.csv', 'w', newline='') as csvfile:
     writer = csv.writer(csvfile, delimiter=',')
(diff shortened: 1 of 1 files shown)


client/fetch-account-history.py
```diff
@@ -19,10 +19,20 @@
...
-end_block = 1204517
+start_block = 784682# 1689688 # 1187284 #781981 #535620
+end_block = 1690155 #660815
 blocks_total = end_block - start_block
 
+def decode_cid(cid_raw):
+    geohash = cid_raw['geohash']
+    digest_raw = cid_raw['digest']
+    if digest_raw.startswith('0x'):
+        digest = b58encode(bytearray.fromhex(digest_raw[2:])).decode("utf-8")
```
pifragile, line 30: i would suggest to always convert to bytes and not the other way around. because for example the digest `0x77f79df7` from LEU as in 
```
  {
    geohash: u0qj9
    digest: 0x77f79df7
  }
```
cannot be decoded to utf-8 properly
image

brenzi: Thats not an issue because we go to base58 anyway. Or is there a possibility that decode to utf8 breaks anything down the line?

pifragile: oh ok, i see, i overlooked the base 58 part. but it is confusing me. to summarize: if we get a utf-8 representation, we leave it as is. and if we get bytes, we encode to base 58 and take the utf-8 representation of that. so we get utf-8 in both cases, but one of them is base58 encoded and one is not. this will cause confusion imo.

i think it would be the safest choice to consistently go with bytes.

brenzi: oh, I overlooked this:
* correct encoding would be kxtkv2md7iP (as list-communities reports it, base58 at the end)
* but my script does kxtkvE8z8 (urf8 at the end)

good catch!

client/fetch-account-history.py
```diff
@@ -29,7 +29,7 @@ def decode_cid(cid_raw):
     if digest_raw.startswith('0x'):
         digest = b58encode(bytearray.fromhex(digest_raw[2:])).decode("utf-8")
     else:
-        digest = digest_raw
+        digest = b58encode(digest_raw).decode("utf-8")
```
brenzi, line 32: now it is consistent, because it encodes to a base58 utf8 string both times, like our rust code does

client/fetch-account-history.py
pifragile: ok it is now more consistent that way. but i have to continue being a pain in the ass, sorry. what is the rationale behind this base58 encoding? i think we are running into a mess of different encodings.
from the substrate side we get two encodings:
- utf-8
- bytes

As can be observed here
image

I think they just display bytes when there is no decoding for utf-8 which is not really consistent but somehow makes sense.
But now intorducing yet another encoding with base58 sounds strange to me, then we have 3 different types of encoding all over the place which will result in a mess with very high probability😅 one example of such a mess would be the situation of comparing a digest with our custom base58 encoding with a digest we get from the polkadot api in utf-8, because they are hard to distinguish.

I also dont understand why `list-communites` uses a base58 encoding. this should also be changed imo for the reasons above.

When simply calling the raw RPC endpoint `encointer_getAllCommunities`, this is what we get:
```
{
    "jsonrpc": "2.0",
    "result": [
        {
            "cid": {
                "geohash": "0x7530716a39",
                "digest": "0x77f79df7"
            },
            "name": "Leu Zurich"
        },
        {
            "cid": {
                "geohash": "0x6b78746b76",
                "digest": "0x45387a38"
            },
            "name": "Aslah"
        }
    ],
    "id": 10
}
```
And this looks consistent to me. If we need a cid string, I would suggest we use `0x6b78746b7645387a38` for Aslah. This leaves no room for misunderstandings.

brenzi: The cid encoding has been defined long ago and in rust it is established. The goal was to have short and recognizable identifiers.  js/apps doesn't support our cid encoding, thats too bad. But js/apps is only for devs anyway. subscan never supported it neither, but we'll replace them with SubQuery anyway.

I see no reason to change cid encoding standard again.

pifragile: ok let's merge then

PR #283, opened 2022-12-08
Encointer-Testnet to run with encointer community-sidechain

echevrier:
- Add Integritee teerex and sidechain pallets
- Add skip-ias-check feature


runtime/Cargo.toml
```diff
@@ -26,6 +26,8 @@ pallet-encointer-ceremonies-rpc-runtime-api = { default-features = false, git =
 pallet-encointer-communities = { default-features = false, git = "https://github.com/encointer/pallets", branch = "master" }
 pallet-encointer-communities-rpc-runtime-api = { default-features = false, git = "https://github.com/encointer/pallets", branch = "master" }
 pallet-encointer-scheduler = { default-features = false, git = "https://github.com/encointer/pallets", branch = "master" }
+pallet-sidechain = { default-features = false, git = "https://github.com/integritee-network/pallets.git", branch = "develop" }
+pallet-teerex = { default-features = false, git = "https://github.com/integritee-network/pallets.git", branch = "develop" }
```
clangenb, line 31: Is this on purpose the develop branch?

clangenb (commented):
I have one question, but it looks good otherwise

runtime/Cargo.toml
echevrier: It was, when I did it ;)
As the pallet mastrer branch is fixed, I will change it

echevrier: Actually, I think polkadot-v.0.9.28 is only working with the develop branch. So we should change it when we do a polkadot upgrade

clangenb (approved):
Looks good to me now!

Issue #278, opened 2022-11-26
Publish docker images of the node and the client


clangenb:
I wanted to use the binaries from here for the integration tests in the encointer-wallet. However, the integration tests there run on MacOS machines, which means our binaries do not work there.

clangenb:
Alternatively, we can also just build for Mac machines in the CI because I am uncertain if our python libs can just execute docker images instead of the binary.

clangenb:
We could also prepare a docker image that contains the python scripts and the encointer-client, so we can bootstrap a community from within the docker image.

clangenb:
@mosonyi thinks docker is easy. So I think we should have two docker images:

1. the encointer-node as usual.
2. The encointer-client. It needs inside:
```
encointer-client-notee
client/py_client
client/bootstrap_demo_community.py
client/cli.py
client/test-data
```
And the entry points should be:

```
[./encointer-client-notee]
[./client/bootstrap_demo_community.py --client encointer-client-notee]
[./client/cli.py --client  encointer-client-notee]
```

clangenb:
Install python deps: https://github.com/encointer/encointer-node/blob/master/scripts/install_python_deps.sh

mosonyi:
@clangenb client is ready to test.
Usage: docker run -it encointer/encointer-client-notee:<version> [encointer-client-notee|bootstrap_demo_community.py|cli.py] <params>
Example:
$ docker run -it encointer/encointer-client-notee:0.0.2 encointer-client-notee --version

mosonyi:
@clangenb the node is added too
example: 
docker run -it encointer/encointer-node-notee:0.0.2 --help

Please edit the Readme accordingly

clangenb:
Done #293.

Thanks a lot, it works like a charm!

PR #286, opened 2022-12-19
Dockerize binaries step 1. #278

mosonyi:
Need to merge to master to be able to test.


clangenb (approved):
Blindly accepted for testing ;)

PR #294, opened 2023-01-09
merge master into community-sidechain branch to get docker feature

echevrier:
needed to build a docker image for this branch


clangenb (approved):
Looks good to me, we just need to do some minor tweaks after wards so that it is obvious that we have a different docker image here.

PR #307, opened 2023-03-11
[cli] do not use utility.batch if only one location added

brenzi:
this is a workaround for https://github.com/encointer/encointer-parachain/issues/169


pifragile (approved):
LGTM

PR #311, opened 2023-04-23
fix typos in bootstrap_demo_community.py

clangenb:


client/bootstrap_demo_community.py
```diff
@@ -285,7 +285,7 @@ def main(ipfs_local, client, url, port, spec_file, test):
     client.await_block(1)
 
     if(not test):
-        print(f"Community {cid} sucessfullly bootatrapped")
+        print(f"Community {cid} successfully bootstrapped")
```
clangenb, line 288: This is the only line I read when running the script. I could no longer bear it XD

PR #325, opened 2023-06-10
[WIP] Polkadot update v0.9.42

### LINKED ISSUE #320: upgrade dependencies to 0.9.42

OverOrion:
Closes #320.


client/src/main.rs
```diff
@@ -174,9 +180,15 @@ fn main() {
...
-                    println!("Metadata:\n {}", Metadata::pretty_format(&meta).unwrap());
+                .runner(|_args: &str, _matches: &ArgMatches<'_>| {
+                    // Most likely upstream bug, because a trait bound is specified, but the struct does not derive it.
+                    // let meta = get_chain_api(matches).metadata();
+                    // let buf = Vec::new();
+                    // let formatter = serde_json::ser::PrettyFormatter::with_indent(b" ");
+                    // let mut ser = serde_json::Serializer::with_formatter(buf, formatter);
+                    // meta.serialize(&mut ser).unwrap();
+                    // let res = String::from_utf8(ser.into_inner());
+                    // println!("Metadata:\n {}", res.unwrap());
```
clangenb: Could also be a missing feature flag.

client/src/main.rs
```diff
@@ -1268,11 +1306,13 @@ fn main() {
                                 Some(sig) => get_pair_from_str(sig),
                                 None => panic!("please specify --signer.")
                             };
-                            let mut api = api.set_signer(signer.clone().into());
+                            let mut api = api;
+                            let signer = ParentchainExtrinsicSigner::new(sr25519_core::Pair::from(signer.clone()));
+                            api.set_signer(signer.clone());
```
clangenb, line 1307: Isn't signer already a `sr25519::Pair`?

runtime/src/lib.rs
```diff
@@ -483,7 +482,7 @@ construct_runtime!(
 		UncheckedExtrinsic = UncheckedExtrinsic,
 	{
 		System: frame_system::{Pallet, Call, Config, Storage, Event<T>} = 0,
-		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Storage} = 2,
+		RandomnessCollectiveFlip: pallet_insecure_randomness_collective_flip::{Pallet, Storage} = 2,
```
clangenb, line 485: Ohhh, nice. I did not know that this existed, I was already worried that we might need to do some magic here.

client/src/utils.rs
```diff
@@ -60,24 +62,28 @@ pub fn get_councillors(api: &Api) -> Result<Vec<AccountId>> {
...
 	api: &Api,
-	xt_hex: String,
+	xt: EncointerXt<C>,
 	tx_payment_cid: Option<&str>,
-) -> Option<H256> {
-	ensure_payment(api, &xt_hex, tx_payment_cid);
-	let tx_hash = api.send_extrinsic(xt_hex, XtStatus::InBlock).unwrap();
+) -> Option<H256>
+where
+	C: Clone + Encode,
```
clangenb: Why do we need the `Clone`?

client/src/utils.rs
```diff
@@ -100,19 +106,20 @@ pub fn contains_sudo_pallet(metadata: &Metadata) -> bool {
...
-		ensure_payment_cc(api, cid_str, xt);
+		ensure_payment_cc(api, cid_str, encoded_xt);
 	} else {
-		ensure_payment_native(api, xt);
+		ensure_payment_native(api, encoded_xt);
 	}
 }
 
-fn ensure_payment_cc(api: &Api, cid_str: &str, xt: &str) {
+fn ensure_payment_cc(api: &Api, cid_str: &str, encoded_xt: &str) {
```
clangenb: Can't we pass a borrow of the unencoded xt?

node/src/service.rs
```diff
@@ -3,9 +3,10 @@
 use encointer_node_notee_runtime::{self, opaque::Block, RuntimeApi};
 use sc_client_api::BlockBackend;
 use sc_consensus_aura::{ImportQueueParams, SlotProportion, StartAuraParams};
+use sc_consensus_grandpa::SharedVoterState;
 pub use sc_executor::NativeElseWasmExecutor;
-use sc_finality_grandpa::SharedVoterState;
-use sc_keystore::LocalKeystore;
+use sc_executor::{HeapAllocStrategy, DEFAULT_HEAP_ALLOC_STRATEGY};
+//use sc_keystore::LocalKeystore;
```
clangenb: Can we remove this one? Btw. here I also do the diff with the node-template runtime when I do the upgrade: https://github.com/paritytech/substrate/tree/master/bin/node-template

clangenb (commented):
Thanks for fighting this war. :)

I see the integration tests failed. Maybe @pifragile can also have a look at it.

client/src/utils.rs
```diff
@@ -60,24 +62,28 @@ pub fn get_councillors(api: &Api) -> Result<Vec<AccountId>> {
...
 	api: &Api,
 	xt: EncointerXt<C>,
 	tx_payment_cid: Option<&str>,
 ) -> Option<H256> {
-	send_xt_hex_and_wait_for_in_block(api, xt.hex_encode(), tx_payment_cid)
+	send_xt_hex_and_wait_for_in_block(api, xt, tx_payment_cid)
 }
 
-pub fn send_xt_hex_and_wait_for_in_block(
+pub fn send_xt_hex_and_wait_for_in_block<C>(
```
haerdib, line 72: If you still wish to have a wrapper function such as that (without a known xt type), instead of `hex` you could use the `opaque` version: `submit_and_watch_opaque_extrinsic_until`. This takes `Bytes` instead of a `String`.

client/src/utils.rs
```diff
@@ -60,24 +62,28 @@ pub fn get_councillors(api: &Api) -> Result<Vec<AccountId>> {
...
+	xt: EncointerXt<C>,
 	tx_payment_cid: Option<&str>,
-) -> Option<H256> {
-	ensure_payment(api, &xt_hex, tx_payment_cid);
-	let tx_hash = api.send_extrinsic(xt_hex, XtStatus::InBlock).unwrap();
+) -> Option<H256>
+where
+	C: Clone + Encode,
+{
+	let encoded_xt = hex::encode(xt.encode());
```
haerdib: To encode an extrinsic into `Bytes`: use `xt.encode().into()` No need to use strings anymore. Or if you want to avoid `Bytes`: Use `xt.encode()` and a `Vec<u8>` and when using  the api-client function, simply use `encoded_xt.into()`.

(check this out: https://github.com/scs/substrate-api-client/blob/master/testing/examples/pallet_transaction_payment_tests.rs#L36-L43)

haerdib (commented):
Looked quickly over this PR and commented on things that hopefully help a little :)

client/src/main.rs
OverOrion: @haerdib Could you please have a look at this? I think this might be an issue, because `Metadata` is defined like this:
``` rust
#[derive(Clone, Debug, Encode, Decode)]
pub struct Metadata {
```
and the `pretty_format` function is:
```rust
#[cfg(feature = "std")]
pub fn pretty_format<Metadata: Serialize>(metadata: &Metadata) -> Option<String> {
```
But the `Serialize` trait is not derived for it, and because of the orphan rule it is not possible for lib users to implement the trait either.

client/src/main.rs
OverOrion: Almost, it is `AppPair` instead of `Pair`.

client/src/utils.rs
OverOrion: Not needed, it was here because I was messing with the extrinsic type, and at one point I needed it to be cloneable. :)

client/src/utils.rs
OverOrion: I wanted to have the borrowing fixed after the CI passed, but unfortunately I had not gotten to that part (yet).

client/src/main.rs
haerdib: Ah, my apologies, that's a real hassle of a naming issue and not very nicely implemented `pretty_format` function. It's only fixed in the later versions (https://github.com/scs/substrate-api-client/pull/532).

The issue here is, that the `Metadata` struct itself does not support Serialize, because that would simply over blow everything as it does contain all metadata plus the maps of events, errors and everything. To serialize the metadata, use the underlying substrate `runtime_metadata`. 

E.g. that should work: 

```rust
 let meta = get_chain_api(matches).metadata().unwrap();
 println!("Metadata:\n {}", Metadata::pretty_format(&meta.metadata).unwrap());
 ```

Comment (OverOrion) [reply client/src/main.rs]: Thanks for the quick answer, fixed it accordingly :+1: 
Also glad to hear it got fixed in a later revision :100:.

**Comment by OverOrion** [comment]
@haerdib First of all thank you for your suggestion, they were very valuable! Currently the CI fails, part of the failure;
```bash
Thread 'tokio-runtime-worker' panicked at 'Bad input data provided to query_fee_details: Could not decode `RuntimeCall::Balances.0`:
Could not decode `Call::transfer::dest`:
Could not decode `MultiAddress`, variant doesn't exist
', runtime/src/lib.rs:566

```

I think I may have messed up upgrading to the newer API client, could you check it?
Also we agreed on with @clangenb and @brenzi that as part of the API client upgrade you could take over, so feel free to push to this branch :)

haerdib:
Sure, I'll have a look at it this afternoon. I won't push on this branch, though - I'd rather create my own (based on your work) if that's okay with you. That way, you can still push some changes if you wish to.

clangenb:
Superseded by #326

PR #334, opened 2023-06-19
Community sidechain branch update

OverOrion:
Commits on master starting with [9338493](https://github.com/encointer/encointer-node/pull/334/commits/933849354e3449f1eb2685322e67642bc6642b66).

Commits starting with 452b1dd1a142350bae1bb128b983ab7f0f9ae6db are missing, I should include them as well, right?


runtime/src/lib.rs
```diff
@@ -356,11 +355,15 @@ impl pallet_balances::Config for Runtime {
...
+	type RuntimeEvent = RuntimeEvent;
 	type DustRemoval = ();
 	type ExistentialDeposit = ConstU128<EXISTENTIAL_DEPOSIT>;
 	type AccountStore = System;
 	type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
+	type HoldIdentifier = ();
+	type FreezeIdentifier = ();
+	type MaxHolds = ConstU32<0>;
+	type MaxFreezes = ConstU32<0>;
 }
```
clangenb, line 367: 🥳

clangenb (approved):
Beautiful, looks good to me!

clangenb:
From my point of view, you can omit those commits for now, as I don't see a benefit for the community-sidechain.

OverOrion:
After a lengthy debugging session with @clangenb we realized that these things are not really documented / straightforward, so added them to the `README`.

README.md
```diff
@@ -2,17 +2,21 @@
 Encointer-node branch to build a Testnet with the encointer [community-sidechain](https://github.com/encointer/community-sidechain).
 This node has the Integritee [teerex](https://github.com/integritee-network/pallets/tree/master/teerex) and [sidechain](https://github.com/integritee-network/pallets/tree/master/sidechain) pallets.
 
+## Notes for developers:
+  - The `my-node-runtime` should point to the `encointer-node` [encointer-node](https://github.com/encointer/encointer-node) or you will have a hard time with decoding things.
+  - The type definitions (`ParentchainExtrinsicParams, ParentchainAdditionalParams, ParentchainSignedExtra`) should should be checked that they are compatible with `pallet_asset_tx_payment` (i.e. `ParentchainPlainTip` changed to `ParentchainAssetTip`).
```
clangenb: I do think that these comments are valuable, but I think they would be even more valuable in the worker's fork. Don't you think?

OverOrion: You are absolutely right, I will move them, thanks! :100:

PR #337, opened 2023-06-21
faucet cli improvements

### LINKED ISSUE #336: [cli] clean up faucet cli
The current state is working, but not convenient:

a sub-command named faucet and then drip,create,dissolve,close would be nicer. But we also need to disambiguate with the existing faucet command which requires no reputation.....so for now I can live with this wild jungle of commands. But:

* [x] better explain cli commands: 
   * close-faucet: lazy garbage collection. can only be called by faucet creator and only once the faucet is empty
   * dissolve-faucet: can only be called by root. moreover, explain arguments: first one is the faucet address, second one is the beneficiary of the remainding funds
* [x] improve create-faucet
   * whitelist should be comma-separated to make it clear that it is a single argument. explain in cli help
   * the return value of this command needs to be the faucet address. First, check extrinsic success, report error, then parse FaucetCreated event for address
* [x] drip-faucet: check extrinsic success and extract error if any
* [x] add list-faucets command.List all faucet accounts, their name, balance, drip amount and whitelist

with the upgrade of the api-client, this should all be easier now. (examples are not there yet, unfortunately: https://github.com/scs/substrate-api-client/issues/591)

pifragile:
*Closes #336

### DIFF
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -172,8 +172,8 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
 			Arg::with_name(WHITELIST_ARG)
 				.takes_value(true)
 				.required(false)
-				.multiple(true)
-				.min_values(1),
+				.value_name("WHITELIST")
+				.help("comma separated list of cids in whitelist"),
 		)
 	}
 
@@ -402,7 +402,7 @@ impl<'a> EncointerArgsExtractor for ArgMatches<'a> {
 	}
 
 	fn whitelist_arg(&self) -> Option<Vec<&str>> {
-		self.values_of(WHITELIST_ARG).map(|c| c.collect())
+		self.value_of(WHITELIST_ARG).map(|s| s.split(",").collect())
 	}
 
 	fn ceremony_index_arg(&self) -> Option<i32> {
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -173,7 +173,7 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
 				.takes_value(true)
 				.required(false)
 				.value_name("WHITELIST")
-				.help("comma separated list of cids in whitelist"),
+				.help("comma separated list of cids in whitelist. no spaces allowed."),
 		)
 	}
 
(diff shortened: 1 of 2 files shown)


client/src/cli_args.rs
```diff
@@ -172,8 +172,8 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
 			Arg::with_name(WHITELIST_ARG)
 				.takes_value(true)
 				.required(false)
-				.multiple(true)
-				.min_values(1),
+				.value_name("WHITELIST")
```
brenzi, line 175: I just realize that people may use spaces after the comma which breaks this argument: maybe indicate in help that no spaces may be used?

client/src/main.rs
```diff
@@ -1657,14 +1657,30 @@ async fn main() {
...
+
+                    let result = api.submit_and_watch_extrinsic_until_success(xt, false);
+
+                    match result {
+                        Ok(report) => {
+                            let events = report
+                            .events
+                            .unwrap();
+
+                            println!("{}", AccountId::decode(&mut events[5].field_bytes()[0..32].as_ref()).unwrap().to_ss58check());
```
brenzi: thread 'main' panicked at 'range end index 32 out of range for slice of length 28', client/src/main.rs:1670:67

brenzi: faucet was created, though

brenzi: image

brenzi (changes requested):

client/src/main.rs
brenzi: I think it is bad practice to use hardcoded indexing (especially for the event index). runtime / pallet changes can break your index assumptions and render this very brittle

brenzi:
all the rest LGTM. thx!

client/src/main.rs
brenzi: you should be able to loop through all events and attempt to decode each into the original EventType RuntimeEvents::FaucetCreated. if successful, extract accountid

brenzi: was an easy fix. committed directly

brenzi (approved):

client/src/main.rs
pifragile: thanks for fixing!

PR #332, opened 2023-06-14
[cli] add list-reputables

haerdib:
closes #315 

I tested it on the Kusama node with the following cmd: 
`  ../target/release/encointer-client-notee -u wss://kusama.api.encointer.org -p 443 list-reputables -v`

The number of the total reputables matches the one of `reputableCount`.

@brenzi  is this what you had somewhat in mind?

Output: 
```                                                                                  
Listing the number of reputables for each community and ceremony for the last 5 cycles
Community ID: u0qj944rhWE
Ceremony ID 39: Total reputables: 26
Ceremony ID 40: Total reputables: 34
Ceremony ID 41: Total reputables: 29
Ceremony ID 42: Total reputables: 29
Ceremony ID 43: Total reputables: 27
Community ID: dpcmj33LUs9
Ceremony ID 39: Total reputables: 2
Ceremony ID 40: Total reputables: 1
Ceremony ID 41: Total reputables: 2
Ceremony ID 42: Total reputables: 2
Ceremony ID 43: Total reputables: 1
u0qj944rhWE,39,5Dz7iJ9F6WFsWjUzbMqSKS2edJrjyp8np7VYy7fVf8SBAfr8
u0qj944rhWE,39,5FeLdean9oRsuxArhmFN2Vj3oQLubRWRQAhTtW3JtTZt1Mp7
u0qj944rhWE,39,5DZrgWMLZXfLpEDm31ziBaABtnfzRUqH67bn86vGaf5X2k58
u0qj944rhWE,39,5DUBgzYP6VuM1zjK2i93AiHQ8vWCPiYUREJj1pzoaTvH4Hnp
u0qj944rhWE,39,5EZQGJmQTiziwmsA7yPYSt391GC68yb8zy3EMsqnsupavX73
[...]
u0qj944rhWE,43,5FCNDbzNgXqW5cYWxvBwS6mwFXPKtKDN5fyQV7s2bsUaPdHS
u0qj944rhWE,43,5DDDjK2HoUxYXuTGCyefhQ5sVf4cWorpx2Pa71xuR14Cv6tw
u0qj944rhWE,43,5HjQdjjztuW7q3Fp275Xo3x7hhLs66EuusP3ep6B4CKpujLy
dpcmj33LUs9,39,5FnSLeCqTtosUzsuG8cWKwKwn8QHhvAL2RsPnSnDj6XXSt1W
[...]

PR #339, opened 2023-07-22
upgrade to polkadot v1.0.0

brenzi:
integrates https://github.com/encointer/pallets/pull/332

CI
* srtool builds with stable now


brenzi:
annoying:
the trait bound `encointer_primitives::scheduler::CeremonyPhaseType: Serialize` is not satisfied

seems we lose the std feature somewhere along the way, but all tomls checked

clangenb:
Serde error is fixed by updating to the newest encointer-pallet's commit, but the build is now stuck in the client because the api-client is still on v0.9.42: https://github.com/scs/substrate-api-client/issues/625

brenzi:
seems like srtool needs an upgrade to a newer rustc

.github/workflows/ci.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           image: paritytech/srtool
           chain: ${{ matrix.runtime }}
           # Can remove that when we can build with stable, see: https://github.com/integritee-network/parachain/issues/200
-          tag: 1.66.1
+          tag: 1.71.0
```
clangenb: This is here for the following reason: By default the newest srtool image is taken. However all images after 1.66.1 only have the stable toolchain installed, which fails as long as we have the nightly in our `rust-toolchain.toml`.

It was substrate that prevented the change to a stable rust-toolchain if I remember correctly, so we can just try switching the toolchain to some stable version, and maybe it works.

node/src/chain_spec.rs
```diff
@@ -188,16 +192,22 @@ fn testnet_genesis(
 			inactivity_timeout: 168500, // 10 years at 30min ceremony cycle
 			meetup_time_offset: 0,
 			endorsement_tickets_per_reputable: 5,
+			_config: Default::default(),
 		},
 		encointer_communities: EncointerCommunitiesConfig {
 			min_solar_trip_time_s: 1, // [s]
-			max_speed_mps: 1,         // [m/s] suggested would be 83m/s,
+			max_speed_mps: 1,
+			_config: Default::default(), // [m/s] suggested would be 83m/s,
```
clangenb: comment is off

clangenb (approved):
Oh nice, the number of changes was turned out to be not so big after all. :)

Approving, and you can merge once you mage the CI pass. Exchanging `nightly-` with `stable-` in the `rust-toolchain.toml` should fix the issue.

.github/workflows/ci.yml
brenzi: it does. thx

node/src/chain_spec.rs
brenzi: good catch 8-)

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2023-01-13" 
+channel = "nightly-2023-05-22"
```
clangenb, line 2: I am actually surprised that it works. Earlier srtool images returned an error because it only has the stable toolchain installed, which conflicted with the one here. Don't know what changed on their end.

PR #342, opened 2023-07-29
fix try runtime

### LINKED ISSUE #340: [CI] should build with try-runtime and runtime-benchmark features as well
currently not tested  (and not working)

brenzi:
closes #340

..and fixes errors in respective build

### DIFF
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -26,6 +26,8 @@ use frame_benchmarking_cli::{BenchmarkCmd, ExtrinsicFactory, SUBSTRATE_REFERENCE
 use sc_cli::{ChainSpec, RuntimeVersion, SubstrateCli};
 use sc_service::PartialComponents;
 use sp_keyring::Sr25519Keyring;
+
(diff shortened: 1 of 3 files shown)


node/src/command.rs
```diff
@@ -26,6 +26,7 @@ use frame_benchmarking_cli::{BenchmarkCmd, ExtrinsicFactory, SUBSTRATE_REFERENCE
 use sc_cli::{ChainSpec, RuntimeVersion, SubstrateCli};
 use sc_service::PartialComponents;
 use sp_keyring::Sr25519Keyring;
+use try_runtime_cli::block_building_info::timestamp_with_aura_info;
```
clangenb, line 31: needs to be feature gated

clangenb (approved):
One minor thing which breaks the build, but looks good to me otherwise

node/src/command.rs
clangenb: fixed it myself

PR #349, opened 2023-09-15
[client] support `AnnouncementSigner` and `CommunityRules` in the spec files

### LINKED ISSUE #348: [cli] support CommunityRules (LoCoFlex) and announcement signer fields
and then 
* [ ] register a LoCoFlex community on Gesell with well-known keys

clangenb:
* Bump pallets for simpler parsing of Bip340 key
* Add minimal Corsican spec file including an `AnnouncementSigner` and `LoCoLight` rules. The bootstrappers are Alice to Ferdie and Alice is the announcement signer.
* Registered Corsica LoCoLight community on Gesell. Closes #348 
* Bump pallets to get refactoring of `LoCoLight` to `LoCoFlex`.


client/test-data/test.locations.mediterranean.locolight.json
```diff
@@ -0,0 +1,55 @@
+{
+  "type": "FeatureCollection",
+  "community": {
+    "meta": {
+      "name": "Mediterranea",
+      "symbol": "MTA",
+      "assets": "QmP2fzfikh7VqTu8pvzd2G2vAd4eK7EaazXTEgqGN6AWoD",
+      "announcementSigner": {"bip340": "0xceda72c19f11aebecf8fb4a0ab3fb2a03e03f283225655baab4728303a560a13"},
+      "rules": "loCoLight"
```
brenzi: I don't like camel case here as it is not intuitive

brenzi (changes requested):
nice and simple. jsut one little thing

client/test-data/test.locations.mediterranean.locolight.json
clangenb: This is substrate standard, we should not go against it: https://github.com/encointer/pallets/blob/1933ce2cc9b196bf8aca13e11c22d8ab254b8a7a/primitives/src/communities.rs#L68

Additionally, if we had better set up the de-/serialization for the other fields, we could simply do: `let cm: CommunityMetdata = serde_json::from_str(&spec["community"]["meta"])`, but we will never be able to do so if we go against the standard here.

clangenb:
I updated the locolight community to Corsica and put `LoCoLight` in the name to make it easily identifyable in the explorer:

The below is taken from a local node, not Gesell.

image

client/test-data/test.locations.corsica.locolight.json
```diff
@@ -0,0 +1,121 @@
...
+      "announcementSigner": {"bip340": "0xceda72c19f11aebecf8fb4a0ab3fb2a03e03f283225655baab4728303a560a13"},
+      "rules": "loCoLight"
+    },
+    "bootstrappers": [
+      "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
+      "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
+      "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
+      "5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy",
+      "5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw",
+      "5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL"
```
clangenb, line 17: Alice to Ferdie

client/test-data/test.locations.corsica.locolight.json
```diff
@@ -0,0 +1,121 @@
+{
+  "type": "FeatureCollection",
+  "community": {
+    "meta": {
+      "name": "Corsica LoCoLight",
+      "symbol": "CLL",
+      "assets": "QmP2fzfikh7VqTu8pvzd2G2vAd4eK7EaazXTEgqGN6AWoD",
+      "announcementSigner": {"bip340": "0xceda72c19f11aebecf8fb4a0ab3fb2a03e03f283225655baab4728303a560a13"},
```
clangenb, line 8: Alice

client/test-data/test.locations.mediterranean.locolight.json
clangenb: Ah sorry, I see now that you don't mean the field name, but the enum variant `loCoLight`. This is also the substrate standard, but I see now that serde can handle both, I change it to UpperCamelCase

clangenb: I was wrong, it doesn't accept it, but we default to `LoCo` when there is an error and I did not notice that. I also fixed the code to panic if the variant does not exist instead of defaulting to `LoCo`.

brenzi (approved):
well, not happy about camelCase as options would exist to change this behavior. But you may be right that we have less trouble like this.
Also I thought we already refactored LoCoLight to LoCoFlex, which we didn't. but as this is an internal value, I don't think its painful to do that later

clangenb:
Oh, I did not know that should have been refactored to `LoCoFlex`. I will do that immediately because it will be a breaking change in the app, which uses deserialization instead of the scale codec. I don't want to have to adjust that later.

PR #354, opened 2023-09-28
Add Corsica LocoFlex community for rococo

clangenb:
Bootstrappers are Alpha, Theta, Zeta


client/test-data/corsica.locoflex.rococo.json
```diff
@@ -0,0 +1,120 @@
...
+      "name": "Corsica LoCoFlex",
+      "symbol": "CLL",
+      "assets": "QmP2fzfikh7VqTu8pvzd2G2vAd4eK7EaazXTEgqGN6AWoD",
+      "announcementSigner": {
+        "bip340": "0xceda72c19f11aebecf8fb4a0ab3fb2a03e03f283225655baab4728303a560a13"
+      },
+      "rules": "loCoFlex"
+    },
+    "bootstrappers": [
+      "5FyWbcwN1TGPdyzRzoEeem3MUcc7jXRs7ZoftZkAQLV47nS7",
```
brenzi, line 14: I'd prefer to add all test accounts (like Adriana community) for flexibility, but I guess the community is already registered? Then we leave it.

client/test-data/corsica.locoflex.rococo.json
```diff
@@ -0,0 +1,120 @@
+{
+  "type": "FeatureCollection",
+  "community": {
+    "meta": {
+      "name": "Corsica LoCoFlex",
+      "symbol": "CLL",
+      "assets": "QmP2fzfikh7VqTu8pvzd2G2vAd4eK7EaazXTEgqGN6AWoD",
+      "announcementSigner": {
+        "bip340": "0xceda72c19f11aebecf8fb4a0ab3fb2a03e03f283225655baab4728303a560a13"
```
brenzi, line 9: It doesn't make much sense to add an announcement signer which isn't a BIP340 / nostr account, because we'llnever be able to sign anything with it. But won't hurt because we can change metadata anytime

client/test-data/corsica.locoflex.rococo.json
```diff
@@ -0,0 +1,120 @@
+{
+  "type": "FeatureCollection",
+  "community": {
+    "meta": {
```
brenzi, line 4: not setting a nominal income will set it to zero and testing will be restricted to accounts with native balance

brenzi: demurrage seems to have a reasonable default value, but setting it explicitly may still be helpful as means of documentation/example

brenzi (approved):

client/test-data/corsica.locoflex.rococo.json
clangenb: To be honest, I did not think to much about it, I simply took the `test-locations-mediterranean.json`, and adapted the existing values, but I agree that these missing values are not ideal.

client/test-data/corsica.locoflex.rococo.json
clangenb: Yeah, but I did not want to put `null` here, as I want to be able to test if the app understands the metadata if the value is set.

client/test-data/corsica.locoflex.rococo.json
clangenb: Yeah, I don't have the other Adriana accounts and its metadata is not in the repository, but I don't really see the benefit of having more than 3 here.

client/test-data/corsica.locoflex.rococo.json
clangenb: Ahh, I did not know that the ceremony rewards are 0 because the default in the chain specs is 1.

clangenb:
I have updated the metadata to include an explicit nominal income and demurrage. I also purged and reregistered the community.

client/test-data/corsica.locoflex.rococo.json
clangenb: Added them

PR #358, opened 2024-02-25
Set MeetupSizeTarget to 15

pifragile:


brenzi:
this PR depends on an outdated version of pallets. need to upgrade node to polkadot-v1.6.0 first #359

brenzi (approved):

PR #361, opened 2024-03-01
upgrade to polkadot v1.8.0

brenzi:
* upgrade all polkadot-sdk dependencies to 1.8.0 off crates.io
* adapt versioning scheme to follow polkadot-sdk with major.minor to be consistent with parachain node
* major refactoring of cli to async
   * only "made it work". didn't optimize for performance yet but there's plenty of potential now to join async ops which don't need to be sequential
   * modularized commands. also here, this is likely only the first step towards a more convenient and maintainable codebase
* upgrade to substrate-api-client 0.17.0. needed to use a patch branch, see 
   * https://github.com/scs/substrate-api-client/issues/745
   * https://github.com/scs/substrate-api-client/issues/748
* builds in `stable` (`fmt` still uses nightly, because stable sucks. fellows runtimes does the same)


brenzi:
CI test bootstrap.. currently fails upon querying fee details:

node:
```
2024-03-05 18:08:48.028 ERROR tokio-runtime-worker runtime: panicked at /home/brenzi/encointer/encointer-node/runtime/src/lib.rs:637:1:
Bad input data provided to query_fee_details: Codec error    
```
client: 
```
thread 'main' panicked at client/src/commands/encointer_core.rs:271:10:
Could not query asset fee details: Client(Call(ErrorObject { code: ServerError(1), message: "Unable to query fee details.", data: Some(RawValue("Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n    0: 0x19c207 - <unknown>!rust_begin_unwind\n    1: 0x87b8 - <unknown>!core::panicking::panic_fmt::hd76cc5a8b1843eaf\n    2: 0x13a198 - <unknown>!TransactionPaymentApi_query_fee_details")) }))
```
guess this is the first attempt to pay fees in CC

clangenb:
Hmmm, this line panicks as it seems: https://github.com/encointer/pallets/blob/79c7f85e45d72a417918a7d5297db1cb0001dcbe/balances-tx-payment/rpc/src/lib.rs#L100

Not sure if it has to do something with the different runtime api versions define here, but the last update to this file is quite old, so I don't think it should matter: https://github.com/paritytech/polkadot-sdk/blob/329c07723614836a05b1d9158e1a8fdea6fb6508/substrate/frame/transaction-payment/rpc/runtime-api/src/lib.rs.

brenzi:
hmm, this here works, so I doubt we have a general issue here:
`../target/release/encointer-client-notee transfer //Alice //Dave 0.123 --cid sqm1v79dF6b --tx-payment-cid sqm1v79dF6b`

it's the whitelist argument which causes the trouble. This here reproduces the issue:
`../target/release/encointer-client-notee transfer //Alice //Dave 0.123 sqm1v79dF6b --cid sqm1v79dF6b --tx-payment-cid sqm1v79dF6b`

ah, I messed up the whitelist argument and dropped the option wrapper

PR #363, opened 2024-03-19
pimp cli

### LINKED ISSUE #282: Client shall support `--atBlock` arg for various cmds.
We can't query stuff for old ceremonies because the data has been purged from the chain after the reputation lifetime. Hence, we should allow the `atBlock` arg to get the state at a previous block, which solves that issue.

We might also want to add a command `getBlockForCeremonyStart`, which returns the block where the transition to that specific ceremony index  happened

brenzi:
closes #282

* make client work with pre-migration `Reputation` type
* make use of `--at` argument for all queries which can work without custom rpc (where block can't be specified)
* make stout parsable as CSV for `list-participants, list-meetups, list-attestees`

### DIFF
--- a/client/src/commands/encointer_ceremonies.rs
+++ b/client/src/commands/encointer_ceremonies.rs
@@ -47,11 +47,11 @@ pub fn list_participants(_args: &str, matches: &ArgMatches<'_>) -> Result<(), cl
 	let rt = tokio::runtime::Runtime::new().unwrap();
 	rt.block_on(async {
 		let api = get_chain_api(matches).await;
-		let at_block = matches.at_block_arg();
+		let maybe_at = matches.at_block_arg();
 		let cid =
-			verify_cid(&api, matches.cid_arg().expect("please supply argument --cid"), at_block)
+			verify_cid(&api, matches.cid_arg().expect("please supply argument --cid"), maybe_at)
 				.await;
-		let current_ceremony_index = get_ceremony_index(&api, at_block).await;
+		let current_ceremony_index = get_ceremony_index(&api, maybe_at).await;
 
 		let cindex = matches.ceremony_index_arg().map_or_else(
 			|| current_ceremony_index,
@@ -70,7 +70,7 @@ pub fn list_participants(_args: &str, matches: &ArgMatches<'_>) -> Result<(), cl
 			println!("Querying {}", registries[i]);
 
 			let count: ParticipantIndexType = api
-				.get_storage_map(ENCOINTER_CEREMONIES, counts[i], (cid, cindex), at_block)
+				.get_storage_map(ENCOINTER_CEREMONIES, counts[i], (cid, cindex), maybe_at)
 				.await
 				.unwrap()
 				.unwrap_or(0);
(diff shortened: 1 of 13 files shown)


client/src/commands/encointer_ceremonies.rs
```diff
@@ -772,7 +909,12 @@ async fn get_attendees_for_community_ceremony(
 	let mut attendees = Vec::new();
 	let mut noshows = Vec::new();
 	for storage_key in storage_keys.iter() {
-		match api.get_storage_by_key(storage_key.clone(), at_block).await.unwrap().unwrap() {
+		match api
+			.get_storage_by_key(storage_key.clone(), at_block)
+			.await
+			.unwrap_or(Some(Reputation::VerifiedLinked(0)))
+			.unwrap()
+		{
```
brenzi: I'm sceptical about this workaround as it could fire back. but it works temporarily to support pre-migration type

client/src/commands/encointer_core.rs
```diff
@@ -44,11 +44,14 @@ pub fn balance(_args: &str, matches: &ArgMatches<'_>) -> Result<(), clap::Error>
 				println! {"{balance:?}"};
 			},
 			None => {
+				if maybe_at.is_some() {
+					panic!("can't apply --at if --cid not set")
+				};
```
brenzi, line 49: It will be tricky to support `--at` here because we make heavy use of custom rpc's

client/src/commands/encointer_ceremonies.rs
```diff
@@ -772,7 +909,12 @@ async fn get_attendees_for_community_ceremony(
 	let mut attendees = Vec::new();
 	let mut noshows = Vec::new();
 	for storage_key in storage_keys.iter() {
-		match api.get_storage_by_key(storage_key.clone(), at_block).await.unwrap().unwrap() {
+		match api
+			.get_storage_by_key(storage_key.clone(), maybe_at)
+			.await
+			.unwrap_or(Some(Reputation::VerifiedLinked(0)))
```
clangenb, line 916: This default seems unintuitive for me, shouldn't it be unverified reputable?

clangenb (approved):
One tiny question, looks good otherwise. I guess you did test it already. :)

client/src/commands/encointer_ceremonies.rs
brenzi: Surprisingly, it only fails in the case of  VerifiedLinked. It's hacky, but this at least preserves the main information of interest

clangenb: Ahhh, ok. I thought that we unwrap an option here, got it now. No, this makes perfect sense to me, actually. All other enum variants haven't changed, so decoding is not a problem. With the custom rpc though, there is just a very high change that there is a `VerifiedUnlinked` in one of the returned reputations, and unfortunately, this makes the whole rpc call fail.

brenzi: added a todo to remove the hack asap
https://github.com/encointer/encointer-node/issues/364

PR #370, opened 2024-06-22
Ab/democracy bots

brenzi:
extending the bot-community.py script to test democracy too:

for each cycle, do
1. submit a new proposal
2. pick a target approval
3. vote with full turnout
4. update proposals whenever

also
* new cli command `export-secret <ACCOUNT>`

the first two cycles will not work because the reputation isn't old enough. ignore


clangenb (approved):
LGTM, I like the implementation for achieving the random target approval rate!

PR #374, opened 2024-09-02
depend on sdk 1.14 and api-client 0.18

brenzi:


brenzi:
suspecting a feature propsagation issue which zepter doesn't fix:
```
 error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
      --> /home/brenzi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/frame-benchmarking-36.0.0/src/v1.rs:1738:2
       |
  1738 | /     sp_runtime::format_runtime_string!(
  1739 | |         "\n* Pallet: {}\n\
  1740 | |         * Benchmark: {}\n\
  1741 | |         * Components: {:?}\n\
  ...    |
  1750 | |         error_message,
  1751 | |     )
       | |_____^ use of undeclared crate or module `alloc`
       |
       = note: this error originates in the macro `sp_runtime::format_runtime_string` (in Nightly builds, run with -Z macro-backtrace for more info)
```

clangenb:
This is an upstream bug, which will be fixed by https://github.com/paritytech/polkadot-sdk/pull/5632, for now I recommend just ignoring the error as it concerns only the benchmarks that we don't care much about in the testnode.

PR #381, opened 2024-12-16
add cli for demurrage proposal

brenzi:


brenzi:
CI fails out of scope, to be resolved with the next dep upgrade https://github.com/encointer/encointer-node/issues/383

PR #385, opened 2025-01-15
Add more commands to the encointer-client docker image and test them in CI

clangenb:
In essence this introduces the feature to run the bot-community, phase.py and faucet.py from the enconter-client docker container. To achieve that, a few minor changes were necessary:

* The `phase.py` CLI has had minor adjustements to be able to run from within a docker container
* The `faucet.py` now takes some arguments and exposes `0.0.0.0` instead of localhost to be reachable from outside the docker
* Breaking: unify the CLI for our scripts in general such that there is only `-u` as a url argument. `-r` has been removed.

Tests:
* The CI does now also run the bootstrap and the bot-community integration tests in docker to ensure that the docker works.

Future work:
* Registering businesses does not work yet, tracking issue: #386

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,277 +26,277 @@ jobs:
       - uses: styfle/cancel-workflow-action@0.11.0
         with:
           access_token: ${{ secrets.GITHUB_TOKEN }}
-  build:
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: Install protoc
-        run: sudo apt-get install protobuf-compiler
-
-        # With rustup's nice new toml format, we just need to run rustup show to install the toolchain
-        # https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659
-      - name: Setup Rust toolchain
-        run: rustup show
-
-      - uses: Swatinem/rust-cache@v2
-        with:
-          shared-key: "release"
-
-      - name: Build
-        run: cargo build --release
-      - name: Upload node
-        uses: actions/upload-artifact@v4
-        with:
-          name: encointer-node-notee-${{ github.sha }}
-          path: target/release/encointer-node-notee
-      - name: Upload CLI client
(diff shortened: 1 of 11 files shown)


client/phase.py
```diff
@@ -22,20 +22,21 @@
 
 global patience
 
+
 @click.command()
-@click.option('-r', '--remote-chain', default=None, help='choose one of the remote chains: gesell.')
-@click.option('--client', default='../target/release/encointer-client-notee', help='Client binary to communicate with the chain.')
+@click.option('--client', default='../target/release/encointer-client-notee',
+              help='Client binary to communicate with the chain.')
+@click.option('-u', '--url', default='ws://127.0.0.1', help='URL of the chain, or `gesell` alternatively.')
```
clangenb, line 29: I know that `-u` and `-r` were not the exact same thing (-r accepts `gesell`) as value and will look up the gesell url. However, I wanted to have a uniform interface. So everything is `-u` know, but I have documented it in help, if `gesell` is also an accepted value.

.github/workflows/ci.yml
```diff
@@ -21,13 +21,13 @@ env:
...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: styfle/cancel-workflow-action@0.11.0
         with:
           access_token: ${{ secrets.GITHUB_TOKEN }}
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
```
clangenb, line 31: Latest is now 24.04, which broke our builds

client/phase.py
brenzi: yeah, that was a hack anyway...

.github/workflows/ci.yml
brenzi: no surprise... :-(

clangenb: TBH the build itself worked, but the docker image environment is a 22.04, and it failed to run there then.

PR #387, opened 2025-01-17
Client updates to fix bootstrapping and bot-community script for parachain

clangenb:
### Main Changes:
* Instead of implicitly deciding that a sudo/collective call should be composed based on the absence of the signer arg, we do now explicitly add a CLI arg to decide if the new-community call should be wrapped or not.
* additionally, we also explicitly check if the transaction should be sent or not via the dry run flag

### There were other smaller fixes necessary for the parachain:
* The tx weight of the parachain is much bigger. I could only batch around 7 add_location calls for a successful proposal. I made this configurable via the `-b` or `--batch-size` CLI argument for now, which defaults to 100 so that it is not a breaking change. Note: Interestingly, the error I got for too many locations was misleading:
   * way too many locations: -> TX would exhaust block limit
   * a little bit too much: -> invalid proposal weight (very misleading error)
* For some reason the faucet amount was off, I could only send around 1 transaction with the `reasonable_balance`. I just added a factor of 100 for now, and decided to not care about it.

Note: I only fixed the `new-community` command. `add-location` should have the same logic in the future with respect to wrapping, batching and dry-running the extrinsic, but I thought this should come with a minor refactoring: https://github.com/encointer/encointer-node/issues/388

I manually tested bootstrapping and the bot-community against the parachain, and it works 🥳.

### DIFF
--- a/client/src/cli_args.rs
+++ b/client/src/cli_args.rs
@@ -38,6 +38,7 @@ const NOMINAL_INCOME_ARG: &str = "nominal-income";
 const DEMURRAGE_HALVING_BLOCKS_ARG: &str = "demurage-halving-blocks";
 const PURPOSE_ID_ARG: &str = "purpose-id";
 const WRAP_CALL_ARG: &str = "wrap-call";
+const BATCH_SIZE_ARG: &str = "batch-size";
 
 pub trait EncointerArgs<'b> {
 	fn account_arg(self) -> Self;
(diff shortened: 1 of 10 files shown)


client/src/cli_args.rs
```diff
@@ -456,6 +463,32 @@ impl<'a, 'b> EncointerArgs<'b> for App<'a, 'b> {
...
+
+	fn wrap_call_arg(self) -> Self {
+		self.arg(
+			Arg::with_name(WRAP_CALL_ARG)
+				.short("w")
+				.long("wrap-call")
+				.takes_value(true)
+				.required(false)
+				.default_value("none")
+				.value_name("none|sudo|collective")
```
brenzi, line 475: that's a nice solution

brenzi (approved):
LGTM

PR #393, opened 2025-01-21
Fix CI for Parachain and refactor a bit

clangenb:
Changes in this PR:
* Provide configurable waiting time after tx, to ensure that they are included in the canonical chain before continuing
* Use that waiting time too to minimize chances of invalid tx's due to a nonce mismatch in case a tx was retracted.
* Some refactoring in the bootstrapping script, which allows running the tests individually. This decreases the CI time greatly, and also reduces the chances of the parachain CI failing due to still occasional invalid tx's.

I think this state is now good enough to integrate in the parachain. Some runs might be flaky, but I think that the tests should work most of the time.


.github/workflows/ci.yml
```diff
@@ -241,7 +241,16 @@ jobs:
...
     strategy:
       matrix:
-        test: [ test_bootstrap_demo_community.sh, test_bot_community.sh, test_register_business.sh ]
+        test: [
+          test_bootstrap_demo_community.sh cc-fee-payment,
+          test_bootstrap_demo_community.sh faucet,
+          test_bootstrap_demo_community.sh reputation-caching,
+          test_bootstrap_demo_community.sh unregister-and-upgrade-registration,
+          test_bootstrap_demo_community.sh endorsement,
+          test_bootstrap_demo_community.sh democracy,
```
clangenb, line 250: Introduced more waiting between the transactions, but we still decreased the CI time ;)

client/bootstrap_demo_community.py
```diff
@@ -85,23 +85,24 @@ def create_community(client, spec_file_path, ipfs_local, signer, wrap_call="none
...
     print(f'📝 Registering Participants for cid: {cid}')
     [client.register_participant(b, cid) for b in accounts]
 
-    blocks_to_wait = 3
     print(f"⌛ Waiting for {blocks_to_wait} blocks, such that xt's are processed...")
     client.await_block(blocks_to_wait)
 
     print(client.list_participants(cid))
     client.next_phase()
+    client.await_block(blocks_to_wait)
```
clangenb, line 100: Just ensure that after every tx there is wait time. I know it is a hassle, but I don't see how we can be sure that nothing happens otherwise.

client/bootstrap_demo_community.py
```diff
@@ -136,236 +139,326 @@ def fee_payment_transfers(client, cid):
...
+
+    print(f'Reputations of account1: {rep}')
+
+    if (f'{cindex -2}', 'sqm1v79dF6b', f'VerifiedLinked({cindex-1})') not in rep:
+        print('Error: Last reputation has not been linked.')
+        exit(1)
+
+    if (f'{cindex-1}', 'sqm1v79dF6b', 'VerifiedUnlinked') not in rep:
+        print('Error: Did not receive reputation.')
         exit(1)
```
clangenb, line 175: This test does now no longer depend on when it is exactly run. It just assumes that there has been one successful ceremony before it is run.

client/bootstrap_demo_community.py
```diff
@@ -136,236 +139,326 @@ def fee_payment_transfers(client, cid):
...
-    client.await_block(2)
+    client.await_block(blocks_to_wait)
     # endorsee count is still 5
     check_participant_count(client, cid, "Endorsee", 5)
 
 
 def balance(x):
     return x * 10 ** 12
 
+def test_first_ceremony_with_early_claim(client, cid, blocks_to_wait):
```
clangenb, line 295: Extracted function from `main`

client/src/commands/encointer_core.rs
```diff
@@ -295,109 +295,113 @@ pub fn apply_demurrage(
 
 async fn listen(matches: &ArgMatches<'_>) {
 	let api = get_chain_api(matches).await;
-	debug!("Subscribing to events");
+
+	let block_count = value_t!(matches, "blocks", u32).ok();
+	let event_count = value_t!(matches, "events", u32).ok();
+
+	wait_for_blocks_or_events(&api, block_count, event_count).await;
+}
```
clangenb, line 303: I extracted this function because I thought I am going to use it elsewhere, but I didn't in the end.

PR #396, opened 2025-02-25
polkadot-stable2024012-2

clangenb:
Notes:
* I did not notice this in the parachain upgrade, but the node and the api-client support now the Extrinsic V5. This is 100% backwards and forwards compatible, so no issues should arise.
* We are in the nirvana when it comes to benchmarking. Some pallets use `()` weights other use weights that we created in 2022. Shall we just completely switch to using `()`? Or we fix the benchmarks, which might be a 1 hour task, see #397.
* I had to downgrade the srtool action, as the build failed for no apparent reason. Now we use the same version as the fellowship. So maybe there is something wrong with the newer version.


client/encointer-api-client-extension/src/extrinsic_params.rs
```diff
@@ -14,11 +14,11 @@ pub type CommunityCurrencyTipExtrinsicParams<T> = GenericExtrinsicParams<T, Comm
 pub type CommunityCurrencyTipExtrinsicParamsBuilder =
 	GenericAdditionalParams<CommunityCurrencyTip, Hash>;
 
-pub type EncointerXt<Call> = UncheckedExtrinsicV4<
+pub type EncointerXt<Call> = UncheckedExtrinsic<
 	ExtrinsicAddress,
 	Call,
 	Signature,
-	GenericSignedExtra<CommunityCurrencyTip, Nonce>,
+	GenericTxExtension<CommunityCurrencyTip, Nonce>,
```
clangenb, line 21: With this update, we implicitly introduce the UncheckedExtrinsicV5. However, it is entirely backwards compatible, and should not introduce any errors. (The parachain will support the V5 too with the stable2412 update).

.github/workflows/ci.yml
```diff
@@ -106,61 +104,27 @@ jobs:
...
-        uses: chevdor/srtool-actions@v0.9.2
+        uses: chevdor/srtool-actions@v0.8.0
         with:
-          image: paritytech/srtool
-          tag: 1.77.0
           chain: ${{ matrix.runtime }}
           runtime_dir: runtime
+          profile: "production"
 
-      - name: Summary
```
clangenb, line 116: Did this while debugging the srtool-build. It turns out that a downgrade is all that was needed XD.

Regardless, we could simplify here a bit as subwasm is included in srtool.

Dockerfile-node
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:22.04
 LABEL description="This is the 2nd stage: a very small image where we copy the Substrate binary."
 
 # Install netcat for websocket healthcheck
-RUN apt update && apt install -yq jq netcat
+RUN apt update && apt install -yq jq netcat ca-certificates
```
clangenb, line 5: Luckily, we test docker in CI too. After the upgrade, the node was complaining about missing root CAs in the docker.

runtime/Cargo.toml
```diff
@@ -61,7 +61,6 @@ sp-offchain = { workspace = true }
 sp-runtime = { workspace = true }
 sp-session = { workspace = true }
 sp-staking = { workspace = true }
-sp-std = { workspace = true }
```
clangenb, line 64: substrate is deprecating sp-std and started to migrate to core and alloc instead.

brenzi (approved):
LGTM, din't check very deeply

PR #401, opened 2025-05-20
Introduce remove-location call

clangenb:
The geo hash in question can be found out with `encointerCommunityLocations(cid, None)`. This give a nice list of locations by geohash. I kind of thought that it is easier to use the storage instead of the custom-rpc in this case. Not sure if it is true.

```
// cid, geohash, location_index
nctr-k remove-location --dryrun --cid s1vrqQL2SD -g s1rgt -l 0
0x3202103e0273317672710fbc30420000d897382798a607000000000000000000e8e91d4883260b00000000000000bc
```

Sanity check:
- [x] I verified that the generated [call](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fsys.ibp.network%2Fencointer-kusama#/extrinsics/decode/0x3202103e0273317672710fbc30420000d897382798a607000000000000000000e8e91d4883260b00000000000000bc) does indeed want to remove the location that I intended.

### DIFF
--- a/client/encointer-api-client-extension/src/communities.rs
+++ b/client/encointer-api-client-extension/src/communities.rs
@@ -5,6 +5,7 @@ use encointer_primitives::{
 	communities::{CidName, CommunityIdentifier, CommunityMetadata, Location},
 };
 use std::str::FromStr;
+use encointer_primitives::communities::GeoHash;
 use substrate_api_client::{ac_compose_macros::rpc_params, rpc::Request, GetStorage};
 
 #[maybe_async::maybe_async(?Send)]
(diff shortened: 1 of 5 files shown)


brenzi:
I'm not sure about using the list index as an argument. I can think of a possible - even if not very likely - failure mode:
1. user A uses PJS to fetch the list of locations and notes the geohash and index
2. user B enacts a removal of another location in the same list with a lower index
3. user A uses the cli to arm the proposal call using the outdated index
4. user A's call will remove the wrong location (or fail if its the last index)

What about using `--lat` and `--lon` arguments for the `remove-location` call? Of course, we'd need to match the arguments against the list of onchain locations with some tolerance for precision and fail upon ambiguity.

If we had these, `add-location` could make use of them as well....(optional)

Moreover, even if we keep the arguments as you propose, I'd like `list-locations` to provide all necessary info (geohash and index per location). Using PJS is pretty awkward for this.

clangenb:
With my limited time this month and considering that we use this once a year - I just wanted to have something working to fix the current setup 😅 I am not sure how fail-safe the matching algorithm is with lon, lat - especially because our communities are loco flex, and the locations are very close to each other.

I have some time towards the end of the month to look at this again with more depth, as there is no hurry now anymore as we removed the location.

However, I think that the wrong index scenario is not really likely, as it means that a location has been removed in the 5 minutes between checking the locations and creating the call.

brenzi (approved):
I created an issue to tackle my comment: #402.
As this PR works as intended and a fix will delay, I suggest we merge it

Issue #406, opened 2026-01-22
[client] add asset proposal features


clangenb:
We need to support the following features:

* Create `SpendAsset`, `IssueSwapAssetOption` proposal
* Get community treasury account id on AssetHub
* ??

PR #407, opened 2026-01-22
Bump to stable2512

### LINKED ISSUE #405: bump encointer-pallets to latest stable2512
We do not test democracy in the CI of the encointer-wallet-app yet, so it is not that urgent to upgrade, however I propose. That we:

* Bump the encointer-pallets to get `spendAsset` and `issueAssetSwapOption`
* Add the pallet-assets to the runtime. The asset id is going to be the same, as for the `foreignAssets` instance on Asset Hub.
* The transfer mechanism is abstracted away in our encointer-treasuries-pallet. So we can simply use our asset instance for spends, instead of sending an XCM

I think this setup is fairly simple, and with this we can keep the encointer pallets runtime config almost the same as production (e.g. same asset id for asset spends.)

Open question: Do we artificially translate the encointer-treasuries account id to the Asset Hub representation, i.e. `Location::new(1, Parachain(1001, EncointerTreasuryAccountId)`?

clangenb:
Closes #405. This simply bumps the PSDK and the encointer-pallets to the latest version. I did not add the Spend/Swap Asset Proposal features in the client yet, but the groundwork has been laid. We follow up in #406.

I noticed that the bash scrip integration tests fail, but this is an issue with IPFS, not with this PR. I suggest fixing this in a separate PR and merging this one as the corresponding docker integration tests work. Tracking issue: #408.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ env:
 jobs:
   cancel_previous_runs:
     name: Cancel Previous Runs
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - uses: styfle/cancel-workflow-action@0.11.0
         with:
@@ -29,8 +29,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Install protoc
-        run: sudo apt-get install protobuf-compiler
+      - name: Install deps
+        run: sudo apt install --assume-yes openssl pkg-config g++ make cmake protobuf-compiler libssl-dev libclang-dev libudev-dev git lz4
 
         # With rustup's nice new toml format, we just need to run rustup show to install the toolchain
         # https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -29,8 +29,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Install protoc
-        run: sudo apt-get install protobuf-compiler
+      - name: Install deps
+        run: sudo apt install --assume-yes openssl pkg-config g++ make cmake protobuf-compiler libssl-dev libclang-dev libudev-dev git lz4
```
clangenb, line 33: I got a missing libclang error. So I simply copied the stuff from the fellowship

Cargo.lock
clangenb, line 1: Many new lines here, but I double-checked that we don't introduce duplicate psdk deps. So all good.

runtime/src/lib.rs
```diff
@@ -576,7 +579,62 @@ impl pallet_encointer_treasuries::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type Currency = pallet_balances::Pallet<Runtime>;
 	type PalletId = TreasuriesPalletId;
+	// Make our live easier by using the same type as in the parachain
+	type AssetKind = VersionedLocatableAsset;
```
clangenb, line 583: This introduces a bit more dependencies, but being aligned with the parachain here will make our life much easier on many ends.

brenzi (approved):

PR #417, opened 2026-02-16
cli for trusted setup

brenzi:


brenzi:
so far, this trusted setup relies on a single party. multiparty MPC coming in.....

